PR #32, opened 2026-01-05
feat: implement V1 streaming to keep subprocess alive across sessions

ShravanSunder:
## Summary

- **Fix 30+ second cold start delay** on every new conversation and session switch
- Implement V1 streaming input mode to keep Claude subprocess alive across the entire plugin lifecycle
- Add pre-warming during plugin load for instant first message response

## Problem

Previously, `resetSession()` (called on "New" button) and `switchSession()` (called when switching conversations) would close and restart the Claude subprocess. Each subprocess spawn takes ~30 seconds, causing significant delays:

```
User clicks "New" → Close subprocess → Spawn new subprocess (30s) → Ready
User switches chat → Close subprocess → Spawn new subprocess (30s) → Ready
```

## Solution

Video of streaming and fast load
https://github.com/user-attachments/assets/91a438ec-8036-4c8e-b34f-142390667ecc

The Claude Agent SDK's V1 streaming mode keeps the subprocess alive when using an `AsyncIterable` prompt. The SDK handles session switching via the `session_id` field in each message - no subprocess restart needed.

**Changes:**
1. **Keep subprocess alive** - `resetSession()` and `switchSession()` no longer close the persistent query
2. **Pre-warm on plugin load** - `preWarm()` spawns subprocess during plugin initialization  
3. **Message channel pattern** - Queue-based async iterable for sending messages to persistent query
4. **Response routing** - Route response chunks to correct handlers for each query

**New behavior:**
```
Plugin loads → Spawn subprocess (30s, background) → Ready
User clicks "New" → Update session state → Instant
User switches chat → Update session state → Instant
Plugin unloads → Close subprocess
```

## Technical Details

- Subprocess spawns once via `setModel()` call in `preWarm()` 
- Session switching handled by SDK via `session_id` field in `SDKUserMessage`
- Config methods (`setModel()`, `setPermissionMode()`, etc.) work on already-running subprocess

## Test plan

- [x] All 1087 tests pass
- [x] TypeCheck passes
- [x] Build succeeds
- [x] Manual testing: First message after reload is fast
- [x] Manual testing: "New" button is instant
- [x] Manual testing: Switching conversations is instant

### DIFF
--- a/src/core/agent/ClaudianService.ts
+++ b/src/core/agent/ClaudianService.ts
@@ -859,27 +859,14 @@ export class ClaudianService {
     this.sessionManager.setSessionId(id, this.plugin.settings.model);
   }
 
-  /**
-   * Switch to a different session WITHOUT restarting the subprocess.
-   * The SDK handles session switching via the session_id field in each message.
-   * This avoids the 30+ second cold start delay on every new conversation.
-   */
+  /** Switches session via session_id in messages, preserving subprocess. */
   async switchSession(newSessionId: string | null): Promise<void> {
-    // Update session manager - next message will use this session_id
     this.sessionManager.setSessionId(newSessionId, this.plugin.settings.model);
-
-    // Clear session-specific state
     this.approvalManager.clearSessionApprovals();
     this.diffStore.clear();
     this.approvedPlanContent = null;
     this.currentPlanFilePath = null;
-
-    // Clear any pending response handlers from previous session
     this.activeResponseResolvers = [];
-
-    // NOTE: We intentionally do NOT close the persistent query here.
-    // The subprocess stays alive and the SDK handles session switching
-    // based on the session_id in each SDKUserMessage.
   }
 
   /** Cleanup resources. */


YishenTu:
Hi, thanks for your effort.

However, I didn't suffer the same issue when i start or switch sessions.

Can you provide more info about it?

ShravanSunder:
@YishenTu There's 2 changes in this PR

1.  fix for loading where it takes 31 seconds to restablish claude here's the video for this taht shows the issue with `master branch`

https://github.com/user-attachments/assets/62e35ac8-dff9-4874-9e30-3d883bdea2ab

2. It adds streaming for claude responses:
Here's vidoe with streaming and fix for process loading
https://github.com/user-attachments/assets/9caff760-7d60-46d2-b677-6eba0783cb28

YishenTu:
## Session Switching Concern

Thanks for this PR! The AsyncIterable approach for keeping the subprocess alive is the right direction. However, I found an issue with the session switching assumption.

### The Problem

The PR assumes that changing `session_id` in `SDKUserMessage` will switch sessions within the same query:

> "SDK handles session switching via session_id field in each message"

I ran an experiment to validate this, and it turns out **this doesn't work**:

- Sending `session_id: ""` does **not** create a new session
- Sending a different `session_id` does **not** switch sessions  
- The session is bound to the query/process and is set at creation time (via `resume` option)

### Impact

This means:
- "New" button appears instant, but Claude still has context from the previous conversation
- Switching conversations doesn't actually load that conversation's history
- Session isolation is broken even though the UI shows different conversations

### What Works

The AsyncIterable approach **does** work well for:
- Eliminating cold start on follow-up messages within the same conversation
- Real-time streaming with `includePartialMessages`

I'll be working on an implementation that uses the AsyncIterable approach for the main chat agent to improve conversation UX. Will share more details soon.

YishenTu:
Done, check PR #46

PR #82, opened 2026-01-11
feat: add Claude Code plugin support

YishenTu:
## Summary

- Add support for discovering and managing Claude Code plugins installed via the CLI
- Plugins are read from `~/.claude/plugins/installed_plugins.json` and filtered by scope
- User-scoped plugins are available in all vaults; project/local plugins only in matching vault
- Settings UI allows enabling/disabling plugins with status indicators (available, unavailable, invalid)
- Plugin slash commands are automatically loaded and namespaced as `plugin-name:command`
- SDK integration passes active plugin configs to the agent

## Test plan

- [ ] Install a Claude Code plugin via CLI and verify it appears in settings
- [ ] Enable/disable plugin and verify state persists across restarts
- [ ] Verify plugin slash commands appear in command dropdown with namespace
- [ ] Test project-scoped plugin only appears in matching vault
- [ ] Verify refresh button reloads plugins, commands, and restarts agent
- [ ] Verify unavailable plugins show error status and are cleaned up on startup

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,10 +90,10 @@ Use it like Claude Code—read, write, edit, search files in your vault.
 ### Features
 
 - **Inline Edit**: Select text + hotkey to edit directly in notes with word-level diff preview
-- **Slash Commands**: Type `/` for custom prompt templates (Settings → Slash Commands)
 - **Instruction Mode**: Type `#` to add refined instructions to system prompt
-- **Skills**: Add `SKILL.md` files to `~/.claude/skills/` or `{vault}/.claude/skills/`
-- **Claude Code Plugins**: Enable plugins via Settings → Claude Code Plugins
+- **Slash Commands**: Type `/` for custom prompt templates (Settings → Slash Commands)
+- **Skills**: Add `SKILL.md` files to `~/.claude/skills/` or `{vault}/.claude/skills/`, recommended to use Claude Code to manage skills
+- **Claude Code Plugins**: Enable plugins via Settings → Claude Code Plugins, recommended to use Claude Code to manage plugins
 - **MCP**: Add external tools via Settings → MCP Servers; use `@mcp-server` in chat to activate
 
 ## Configuration


YishenTu:
@codex

YishenTu:
@codex ignore the @ mention feature, it was unrelated change that has been resolve, review the rest pr

YishenTu:
### Code review

Found 2 issues:

1. Version comparison uses string comparison instead of semantic versioning. For versions like "10.0.0" vs "2.0.0", string comparison incorrectly returns "2.0.0" as newer because "1" < "2" lexicographically. This will cause the wrong plugin version to be selected when multiple versions with different digit lengths coexist.

https://github.com/YishenTu/claudian/blob/3e92b186719d4451c7d6f625822ad6cf0bc3f718/src/core/plugins/PluginStorage.ts#L114-L117

2. `getPluginCommandPaths()` returns `plugin.installPath` instead of `plugin.pluginPath`. For marketplace plugins with a `source` field, these are different directories (pluginPath = installPath + source subdirectory). This will cause plugin commands to fail to load for any marketplace plugin that uses the source field.

https://github.com/YishenTu/claudian/blob/3e92b186719d4451c7d6f625822ad6cf0bc3f718/src/core/plugins/PluginManager.ts#L166-L172

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

YishenTu:
**Update**: After investigating the actual plugin directory structure in `~/.claude/plugins/`, issue #2 (installPath vs pluginPath) is a **false positive**:

- Plugins with `commands/` directories have `installPath` pointing directly to the correct location
- Marketplace plugins use `source: "./"`, so `pluginPath == installPath`
- The Claude CLI pre-computes the correct path when installing

Only issue #1 (string-based version comparison) remains valid.

PR #208, opened 2026-01-29
feat: add retry button for last assistant message

irmowan:
Add a retry button that appears on hover over the last assistant message. Clicking it removes the current response and resends the user message.

**This is very helpful as I often met 500 internal server issue and wants to retry**.

- Add retry callback mechanism to MessageRenderer
- Add retryFromMessage method to InputController
- Show retry button only on last assistant message (both streaming and history)
- Add CSS styles for hover-visible retry button

Original:
image

Update: (with interrupt)
image
Update:
image

### DIFF
--- a/tests/unit/features/chat/controllers/InputController.test.ts
+++ b/tests/unit/features/chat/controllers/InputController.test.ts
@@ -105,6 +105,8 @@ function createMockDeps(overrides: Partial<InputControllerDeps> = {}): InputCont
       addMessage: jest.fn().mockReturnValue({
         querySelector: jest.fn().mockReturnValue(createMockEl()),
       }),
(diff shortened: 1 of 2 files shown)


YishenTu:
retry and rewind are the features i am planing to add, still considering the proper approach, may merge this pr in the future.

irmowan:
> retry and rewind are the features i am planing to add, still considering the proper approach, may merge this pr in the future.

Good to know, rewind can also including edit features. Your project is very awesome and I am very gald to help contribute some features.

YishenTu:
`rewind` added it #224, note that rewind only works in new session, previous session does not have checkpoint file hence not possible to rewind.

but for `retry`, since i don't normally encounter api error issues, it's still under consideration how to implement this feature.

PR #245, opened 2026-02-06
feat: add switch-to-tab-N commands for quick tab navigation

Feng-H:
Add 9 new commands (switch-to-tab-1 through switch-to-tab-9) that allow users to quickly switch between Claudian tabs using keyboard shortcuts.

Commands only appear when the target tab exists (e.g., switch-to-tab-3 won't show if there are only 2 tabs open).

Users can bind these to hotkeys like Cmd+1-9 or Fn+1-9 in Obsidian's hotkey settings.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -181,6 +181,31 @@ export default class ClaudianPlugin extends Plugin {
       },
     });
 
+    // Switch to tab 1-9 commands
+    for (let n = 1; n <= 9; n++) {
+      this.addCommand({
+        id: `switch-to-tab-${n}`,
+        name: `Switch to tab ${n}`,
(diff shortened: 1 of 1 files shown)


YishenTu:
### Code Review Findings

1. **[P2] Tab switch commands target first pane instead of active pane** (`src/main.ts:190`)

The new `switch-to-tab-N` commands always resolve `getLeavesOfType(...)[0]`. In a multi-pane workspace, triggering the hotkey from pane B can switch pane A's tab, which breaks expected "current pane" navigation behavior. Resolve the active Claudian view first (active leaf/view), and only fallback to the first leaf when no active Claudian pane exists.

PR #263, opened 2026-02-10
fix: entries is not iterable

mattshma:
开启插件时报错：

image

提示："claudian" 加载失败。修复后编译测试能正常开启插件。

### DIFF
--- a/src/core/plugins/PluginManager.ts
+++ b/src/core/plugins/PluginManager.ts
@@ -93,7 +93,8 @@ export class PluginManager {
       for (const [pluginId, entries] of Object.entries(installedPlugins.plugins)) {
         if (!entries || entries.length === 0) continue;
 
-        const entry = selectInstalledPluginEntry(entries, normalizedVaultPath);
+        const entriesArray = Array.isArray(entries) ? entries : [entries];
+        const entry = selectInstalledPluginEntry(entriesArray, normalizedVaultPath);
         if (!entry) continue;
 
         const scope: PluginScope = entry.scope === 'project' ? 'project' : 'user';


YishenTu:
Thanks for the PR! I looked into this and have some questions before we can merge.

The crash happens at `selectInstalledPluginEntry` (line 52) where `for (const entry of entries)` throws "entries is not iterable" — meaning the value in your `installed_plugins.json` is a single object instead of an array.

However, the SDK consistently writes arrays to `~/.claude/plugins/installed_plugins.json`. The type definition (`InstalledPluginsFile`) declares `plugins: Record<string, InstalledPluginEntry[]>`, and on our end every entry is a proper array.

Could you share the contents of your `~/.claude/plugins/installed_plugins.json`? That would help us understand:
1. Whether an older SDK version wrote a different format
2. Whether the file was hand-edited or corrupted
3. Whether a defensive fix is actually warranted, or if the root cause is elsewhere

If it turns out to be a one-off corrupted file, fixing the file itself is the right solution. If it's a real format variant from a specific SDK version, we'd want to handle it properly (including updating the type definition).

mattshma:
`installed_plugins.json` content:
```json
{
  "version": 2,
  "plugins": {
    "project-management-suite@claude-code-templates": [
      {
        "scope": "user",
        "installPath": "/Users/x/.claude/plugins/cache/claude-code-templates/project-management-suite/unknown",
        "version": "unknown",
        "installedAt": "2025-11-04T02:28:47.632Z",
        "lastUpdated": "2025-11-04T02:28:47.632Z",
        "gitCommitSha": "c4ad408b77a73fd6c53eb5a3750bbce8e76184a4"
      }
    ]
  }
}
```

`claudecode` was installed use `pnpm`.

YishenTu:
Thanks for sharing the sample `installed_plugins.json` in the thread. That sample is valid (each plugin maps to an array), so the root cause is likely not that exact snapshot by itself.

What we can confirm from code:
- The crash site is `for (const entry of entries)` in `selectInstalledPluginEntry` (`src/core/plugins/PluginManager.ts`).
- That means `entries` was non-iterable at runtime in the failing environment.

So the most likely issue is runtime shape drift at the file boundary (older/stale/corrupted state, cross-version data, or another runtime path), even if the currently posted file is correct.

Suggestions:
1. Keep the defensive normalization (`Array.isArray(entries) ? entries : [entries]`) to avoid hard crashes on external persisted data.
2. Normalize before any `entries.length` checks.
3. Add a regression test in `tests/unit/core/plugins/PluginManager.test.ts` for the non-array `plugins[pluginId]` case.
4. Align types/runtime expectations (either explicit runtime validator or widen `InstalledPluginsFile` to reflect tolerated input).

If possible, please also share the original stack trace/log from the failing run so we can confirm this is the exact runtime path that produced the error.

eryajf:
I have encountered the same problem. If there is any information that needs to be provided for cooperation, I can provide it. Currently, when the plugin is enabled, an error will be reported

image

YishenTu:
The `installed_plugins.json` spec expects `plugins[pluginId]` to be an array, but I've merged this to loosen the format restriction so a crash won't happen during initialization if it's malformed. A follow-up patch also adds an Obsidian `Notice` when normalization is triggered, so users are informed that their installed plugins file has a malformed entry rather than silently masking it.

eryajf:
@mattshma  Thank you for your contribution, @YishenTu  thank you for merging, now it can works

PR #264, opened 2026-02-10
feat: configurable Claude home directory and runtime model detection

pennn778:
## Summary

- **Configurable Claude home directory**: Replace all hardcoded `.claude` path references with a centralized `claudePaths` module that reads a configurable directory name. This allows users of custom CLI builds (e.g., `claude-internal`) that store data in `~/.claude-internal/` to configure the plugin accordingly. Includes settings UI, i18n support for all 10 locales, and 15 unit tests.
- **Auto-detect available models from SDK at runtime**: Replace the hardcoded model list with runtime detection via `persistentQuery.supportedModels()`, so `claude-internal` users see their actual available models. Priority: SDK models > env-var models > DEFAULT_CLAUDE_MODELS.
- **Documentation updates**: Update CLAUDE.md, README, and add Chinese translation (README_CN.md).

## Changes

- Add `src/utils/claudePaths.ts` with get/set functions for the configurable dir name
- Convert static path constants in all 8 storage modules to dynamic functions
- Update PluginManager, AgentManager, sdkSession, and path utils to use dynamic paths
- Add `ClaudianService.getSupportedModels()` to fetch models from SDK
- Update `ModelSelector` to prefer SDK-detected models at runtime
- Add settings UI for Claude home directory (persisted to `data.json` to avoid bootstrap circularity)
- Bootstrap dir name in `main.ts` before loading settings
- Add i18n keys to all 10 locales
- Add unit tests for claudePaths, getSupportedModels, and InputToolbar SDK model integration

## Test plan

- [x] Unit tests pass
- [x] Verify plugin loads with default `.claude` directory
- [x] Change Claude home directory in settings to `.claude-internal` and verify storage paths update
- [x] Verify model selector shows SDK-detected models when service is ready
- [x] Verify fallback to env-var/default models when SDK models unavailable


YishenTu:
PR review findings:

1. **Critical**: `setClaudeHomeDirName` is called on every keystroke
   - File: `src/features/settings/ClaudianSettings.ts:626`
   - `TextComponent.onChange` fires per-keystroke, but this mutates global path resolution immediately. Intermediate values (e.g. `.c`, `.cl`) can become active while user is typing.
   - Suggested fix: in settings UI, only persist value to `data.json` and show restart notice; avoid calling `setClaudeHomeDirName()` here.

2. **Critical**: bootstrap path applies persisted `claudeHomeDirName` without validation
   - File: `src/main.ts:66` and `src/main.ts:67`
   - `data.json` is trusted directly; malformed/manual values can flow into global path resolution.
   - Suggested fix: validate before calling setter (or enforce validation inside setter).

3. **High**: mutation boundary has no validation
   - File: `src/utils/claudePaths.ts:6`
   - `setClaudeHomeDirName(dirName: string)` currently accepts arbitrary strings.
   - Suggested fix: enforce validation in setter itself (`startsWith('.')`, no `/` or `\\`, reject `.` and `..`, reject empty).

4. **Important**: current validation allows `.` and `..`
   - File: `src/features/settings/ClaudianSettings.ts:618`
   - Current guard blocks separators but still allows `.` and `..`.
   - Suggested fix: reject both explicit values in the validation condition.

5. **High**: hardcoded command path remains in delete flow
   - File: `src/core/storage/SlashCommandStorage.ts:52`
   - `delete()` still uses `COMMANDS_PATH` instead of `getCommandsPath()`, so configurable Claude dir migration is incomplete for this code path.
   - Suggested fix: replace with dynamic path helper for consistency with the rest of the migration.

pennn778:
## Review feedback addressed

Pushed a fix commit (`7ca68e1`) addressing all 5 review comments:

### Changes

1. **Critical — setter called on every keystroke** (comment #1)
   - Settings UI no longer calls `setClaudeHomeDirName()` in `onChange`. Only persists the value to `data.json`; a restart is required for path changes to take effect.

2. **Critical — bootstrap without validation** (comment #2)
   - `setClaudeHomeDirName()` now internally validates via `isValidClaudeHomeDirName()`. Invalid values from `data.json` are silently ignored (default `.claude` remains).

3. **High — setter has no validation** (comment #3)
   - Added `isValidClaudeHomeDirName()` that rejects: empty strings, `.`, `..`, names without leading `.`, and names containing path separators (`/`, `\`).

4. **Important — validation allows `.` and `..`** (comment #4)
   - Settings UI validation now uses the shared `isValidClaudeHomeDirName()` which rejects both `.` and `..`.

5. **High — hardcoded `COMMANDS_PATH` in delete flow** (comment #5)
   - `SlashCommandStorage.delete()` now uses `getCommandsPath()`. The deprecated `COMMANDS_PATH` constant has been removed.

### Verification

- All 4454 tests pass
- TypeScript type check passes
- Lint passes
- Build succeeds

PR #275, opened 2026-02-14
fix: lock focus during streaming, release after response completes

w65467991:
## Summary

- Focus indicator (current note chip) was permanently locked after the first message because `handleFileOpen` was gated on `sessionStarted`, which never resets during a conversation
- Now the lock is based on streaming state: file-open events are skipped while `isStreaming` is true in `ClaudianView` and resume once the response finishes
- During an active session (between messages), the focus chip follows file switches without clearing manual @-mention attachments

## Test plan

- [ ] Open sidebar, switch files before sending any message — focus chip should follow
- [ ] Send a message, while AI is responding switch files — focus chip should stay locked
- [ ] After AI finishes responding, switch files — focus chip should follow again
- [ ] Verify @-mentioned files are not cleared when switching files mid-session

### DIFF
--- a/src/features/chat/ClaudianView.ts
+++ b/src/features/chat/ClaudianView.ts
@@ -511,12 +511,13 @@ export class ClaudianView extends ItemView {
       this.plugin.app.vault.on('modify', markDirty)
     );
 
-    // File open event
+    // File open event — locked while streaming, released after
     this.registerEvent(
       this.plugin.app.workspace.on('file-open', (file) => {
-        if (file) {
-          this.tabManager?.getActiveTab()?.ui.fileContextManager?.handleFileOpen(file);
-        }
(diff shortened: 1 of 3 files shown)


YishenTu:
Current Note Chip locked after session start is by design. The purpose of it is to link a session with a note, which is recommended.

If you reference another note mid session, the record is kept in chat history and Claudian knows that.

Here's a little example of how Claudian handle ref:

```
// First round of interaction
Summarize this note.
<current_note>
Note-A.md
</current_note>
{first AI Response}

// Second round of interaction 
Compare to @Note-B.md
{second AI response}

// Following interactions
...
```

As you can see, all referenced notes are kept in chat history, and the chip is solely an indicator of the first ref note.

PR #285, opened 2026-02-24
fix: bypass CORS for MCP HTTP/SSE server testing

execsumo:
## Problem

The MCP server verification button in Obsidian settings fails with CORS errors when testing **any** HTTP/SSE server (local or remote). This happens because Obsidian runs in Chromium's renderer process, which sends an `Origin: app://obsidian.md` header with every `fetch`/`EventSource` request and enforces strict CORS policies. Since MCP servers are designed for CLI tools like Claude Code (pure Node.js, no CORS), they don't whitelist `app://obsidian.md` in their `Access-Control-Allow-Origin` response — causing all HTTP/SSE server verification to fail.

This affects both local and remote MCP servers equally. The actual agent execution (via Claude Code CLI) is unaffected since it runs in Node.js where CORS doesn't apply.

## Solution

Replace browser-based `SSEClientTransport` and `StreamableHTTPClientTransport` with native Node.js `http`/`https` modules for MCP server verification. Since Obsidian's Electron environment has full Node.js integration, we can use Node's native HTTP client which never sends an `Origin` header and completely bypasses browser CORS restrictions.

### Changes

- **HTTP type**: Direct POST with JSON-RPC `tools/list` request, parse JSON response
- **SSE type**: Open GET stream, wait for `endpoint` event, then POST to that endpoint
- **stdio type**: Unchanged (already uses Node.js child process)

## Testing

- Verified SSE server testing works against a server without CORS headers
- Verified HTTP server testing works against a server without CORS headers
- `npm run typecheck` passes
- `npm run build` succeeds

### DIFF
--- a/src/core/mcp/McpTester.ts
+++ b/src/core/mcp/McpTester.ts
@@ -1,5 +1,7 @@
 import { Client } from '@modelcontextprotocol/sdk/client';
+import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse';
 import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio';
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp';
 import * as http from 'http';
 import * as https from 'https';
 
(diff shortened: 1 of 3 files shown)


execsumo:
Reworked the code and checked errors, the ones that are left were there in the main branch before the fork.

YishenTu:
Pushed updates to this PR branch to preserve the CORS-bypass goal while keeping MCP protocol handling on official SDK transports.

What was done:
- Kept StdioClientTransport unchanged.
- Restored SDK remote transports:
  - SSEClientTransport for type: sse
  - StreamableHTTPClientTransport for type: http
- Added a Node-backed fetch adapter (via http/https) and passed it into SDK transports so verification bypasses renderer CORS.
- Kept header forwarding via requestInit.headers.
- Updated MCP unit/integration tests to assert remote transports receive the custom fetch and configured headers.

Validation run locally:
- npm run typecheck
- npm run lint
- npm run test -- --selectProjects unit --testPathPatterns tests/unit/core/mcp/McpTester.test.ts
- npm run test -- --selectProjects integration --testPathPatterns tests/integration/core/mcp/mcp.test.ts

Commit pushed: ad6dc36d

PR #295, opened 2026-02-28
feat: load global user-level skills and MCP servers

execsumo:
- SkillStorage: load skills from ~/.claude/skills/ alongside vault skills (.claude/skills/). Vault skills take precedence on name collision. Global skills get id prefix 'skill-global-'. Isolated vault-loading errors so global skills load even if the vault directory doesn't exist yet.

- McpStorage: load MCP servers from ~/.claude/settings.json alongside vault .claude/mcp.json. Vault servers take precedence on name collision. Restructured load() to isolate vault try/catch from global merge.

- AgentManager already loads global agents from ~/.claude/agents - no changes needed.

- Added unit tests for both global loading paths (5 new SkillStorage tests, 6 new McpStorage tests) with fs mocks.

### DIFF
--- a/src/core/storage/SkillStorage.ts
+++ b/src/core/storage/SkillStorage.ts
@@ -17,30 +17,32 @@ export class SkillStorage {
     const loadedNames = new Set<string>();
 
     try {
-      const folders = await this.adapter.listFolders(SKILLS_PATH);
+      if (await this.adapter.exists(SKILLS_PATH)) {
+        const folders = await this.adapter.listFolders(SKILLS_PATH);
 
-      for (const folder of folders) {
-        const skillName = folder.split('/').pop()!;
-        const skillPath = `${SKILLS_PATH}/${skillName}/SKILL.md`;
+        for (const folder of folders) {
+          const skillName = folder.split('/').pop()!;
+          const skillPath = `${SKILLS_PATH}/${skillName}/SKILL.md`;
 
-        try {
-          if (!(await this.adapter.exists(skillPath))) continue;
+          try {
+            if (!(await this.adapter.exists(skillPath))) continue;
 
-          const content = await this.adapter.read(skillPath);
-          const parsed = parseSlashCommandContent(content);
+            const content = await this.adapter.read(skillPath);
+            const parsed = parseSlashCommandContent(content);
 
-          skills.push(parsedToSlashCommand(parsed, {
-            id: `skill-${skillName}`,
-            name: skillName,
-            source: 'user',
-          }));
-          loadedNames.add(skillName);
-        } catch {
-          // Non-critical: skip malformed skill files
+            skills.push(parsedToSlashCommand(parsed, {
+              id: `skill-${skillName}`,
+              name: skillName,
+              source: 'user',
+            }));
+            loadedNames.add(skillName);
+          } catch {
+            // Non-critical: skip malformed skill files
+          }
         }
       }
     } catch {
-      return skills;
+      // Non-critical: skip vault skills if directory missing or inaccessible
     }
 
     // Also load user-level skills from ~/.claude/skills (global Claude Code skills).


YishenTu:
> SkillStorage: load skills from ~/.claude/skills/ alongside vault skills (.claude/skills/). Vault skills take precedence on name collision. Global skills get id prefix 'skill-global-'. Isolated vault-loading errors so global skills load even if the vault directory doesn't exist yet.

i do think we already support global skill discovery.

YishenTu:
> @YishenTu That's true for Agents and Plugins, but I think Skills and MCP were only loading from the directory for the Obsidian Vault ".claude", but not the global user directory "~/.claude". I verified this on two machines, a Windows PC and a Macbook Pro, and neither loaded the user-level Skills and MCPs.
> 
> Still possible that it was just my instances, but from the code it does look like Skills and MCPs were just referencing the .claude in the vault?

to be more precise, Claude Agent SDK is handling the skill discovery. we only parse what's inside vault/skills into the setting page. the consideration behind this design is that global skills usually contains coding-related skills, which is not suitable for obsidian operation.

if you ask claudian, "list all your skills", you will see local as well as global skills are dicoverable.

execsumo:
> > SkillStorage: load skills from ~/.claude/skills/ alongside vault skills (.claude/skills/). Vault skills take precedence on name collision. Global skills get id prefix 'skill-global-'. Isolated vault-loading errors so global skills load even if the vault directory doesn't exist yet.
> 
> i do think we already support global skill discovery.

@YishenTu  That's true for Agents and Plugins, but I think Skills and MCP were only loading from the directory for the Obsidian Vault ".claude", but not the global user directory "~/.claude". I verified this on two machines, a Windows PC and a Macbook Pro, and neither loaded the user-level Skills and MCPs.

Maybe that was typo / vibe code slip and you wanted it to reference user-level configurations in ~/.claude for Skills, MCPs, Agents, and Plugins? I think having only user-level tools would work too if you want to keep the code simpler, but only having vault/project-level tools means users have to copy the tools between vaults if they want to use the same tools.

execsumo:
> > @YishenTu That's true for Agents and Plugins, but I think Skills and MCP were only loading from the directory for the Obsidian Vault ".claude", but not the global user directory "~/.claude". I verified this on two machines, a Windows PC and a Macbook Pro, and neither loaded the user-level Skills and MCPs.
> > Still possible that it was just my instances, but from the code it does look like Skills and MCPs were just referencing the .claude in the vault?
> 
> to be more precise, Claude Agent SDK is handling the skill discovery. we only parse what's inside vault/skills into the setting page. the consideration behind this design is that global skills usually contains coding-related skills, which is not suitable for obsidian operation.
> 
> if you ask claudian, "list all your skills", you will see local as well as global skills are dicoverable.

ah ok, that makes sense. I'll go back to earlier version and try "list all your skills". Thanks Yishen. Love the tool, BTW.

YishenTu:
SDK is handling skill discoverability, and it follows Claude Code's pattern, loading from ~/.claude/skills/ and project/.claude/skills/ , it cant be turned off or modified.

what we did is that in setting page, we list commands/skills in project/vault level, which is much more related to obsidian operation, to allow user to be able to modify them directly in the UI.

for global skills (~/.claude/skills), just ask Claudian, "list all your skills" or "use this skills to do foo" , it can load them automatically without any settings.

execsumo:
gotcha - maybe we can add a option in the settings to show or hide global tools 🤔 
that way, users like me who want / need global tools can just toggle and the setting stays persistent vs asking to list and enable all tools in each conversation instance.

YishenTu:
> gotcha - maybe we can add a option in the settings to show or hide global tools 🤔 that way, users like me who want / need global tools can just toggle and the setting stays persistent vs asking to list and enable all tools in each conversation instance.

i do want to isolate vault and user skills/commands/agents to be seen and modified within Claudian, since vault operation is very different from coding/other tasks, however, you can use them without any changes, this is mostly SDK's behaviour.

MCP on the other hand, since i dont use them very much, i have no tendency towards them. but the general principle is the same, user should be able to use it, but not touch it.

PR #297, opened 2026-02-28
fix(i18n): Fix and complete the zh_CN text!

Peingxious:
Changed some localized content to full localization, supplemented it twice to ensure completeness (only some prompts are not localized as they were not found in i18n)


YishenTu:
already fix i18n

PR #323, opened 2026-03-06
fix(cli-detection): resolve nvm-installed Claude CLI in GUI apps

Michael-OvO:
## Problem

macOS GUI apps like Obsidian don't source shell profiles (`~/.zshrc`, `~/.bashrc`), so environment variables set by nvm — specifically `NVM_BIN` — are never available at runtime. The existing nvm detection in `getExtraBinaryPaths()` relied solely on `process.env.NVM_BIN`, causing **"Claude CLI not found"** errors for users who installed Claude Code via npm under nvm.

This is the most common Node version manager on macOS/Linux ([70%+ market share](https://2023.stateofjs.com/en-US/other-tools/#javascript_flavors)), so this affects a significant number of users.

### Root cause

There are two independent search paths for binary detection:

| Function | Purpose | nvm handling |
|----------|---------|-------------|
| `getExtraBinaryPaths()` (env.ts) | Find Node.js binary | `NVM_BIN` env var only |
| `findClaudeCLIPath()` (path.ts) | Find Claude CLI binary | No nvm support |

Both paths failed in GUI apps because `NVM_BIN` is never set outside a shell session.

### How other version managers are handled

The codebase already handles this for other tools with **hardcoded fallback paths**:

- **Volta**: `~/.volta/bin` (fixed path, works in GUI apps)
- **asdf**: `~/.asdf/shims` (fixed path, works in GUI apps)
- **fnm**: `~/.fnm` + env vars (partially works)

nvm is harder because its binary path includes a version number (`~/.nvm/versions/node/v22.18.0/bin/`), requiring filesystem resolution rather than a static path.

## Solution

Added `resolveNvmDefaultBin()` which resolves nvm's default Node version from the filesystem:

1. Read `~/.nvm/alias/default` (or `$NVM_DIR/alias/default`)
2. Follow alias chains if needed (e.g., `default` → `lts/*` → `lts/jod` → `v22.18.0`)
3. Match the resolved version against installed directories in `~/.nvm/versions/node/`
4. Return the `bin/` directory of the highest matching version

This fallback is used in **both** detection paths:
- `getExtraBinaryPaths()` — so Node.js can be found for JS-based CLI files
- `findClaudeCLIPath()` — so the `claude` binary itself can be found

### Design choices

- **Alias chain resolution** with depth limit (5) prevents infinite loops from circular aliases
- **Highest version match** when multiple versions match a major alias (e.g., `22` → `v22.18.0` over `v22.5.0`)
- **Respects `NVM_DIR`** env var when set, falls back to `~/.nvm`
- **Safe no-op** when nvm is not installed — `readFileSync` on a missing file throws, caught and returns `null`
- **`NVM_BIN` still takes priority** — the filesystem fallback only runs when the env var is absent

## Changes

| File | Change |
|------|--------|
| `src/utils/path.ts` | Added `resolveNvmAlias()` and `resolveNvmDefaultBin()`, integrated into `findClaudeCLIPath()` |
| `src/utils/env.ts` | Imported `resolveNvmDefaultBin`, added fallback in `getExtraBinaryPaths()` |
| `tests/unit/utils/env.test.ts` | 4 new tests: basic resolution, `NVM_DIR` support, multiple version matching, alias chain following |
| `tests/unit/utils/path.test.ts` | 1 new test: `findClaudeCLIPath` finds Claude via nvm resolution |

## Test plan

- [x] All 4494 existing tests pass
- [x] 5 new tests cover: simple alias, `NVM_DIR` override, multi-version matching, `lts/*` alias chains, CLI path detection
- [x] Typecheck clean
- [x] Lint clean
- [x] Build succeeds
- [x] Manually verified on macOS with nvm — Obsidian plugin detects Claude CLI after this change

### DIFF
--- a/src/utils/path.ts
+++ b/src/utils/path.ts
@@ -293,11 +293,30 @@ function getNpmCliJsPaths(): string[] {
  * Resolves an nvm alias to a version string by following the alias chain.
  * e.g., "default" → "lts/*" → "lts/jod" → "v22.18.0" → "22"
  */
+const NVM_LATEST_INSTALLED_ALIASES = new Set(['node', 'stable']);
+
+function isNvmBuiltInLatestAlias(alias: string): boolean {
+  return NVM_LATEST_INSTALLED_ALIASES.has(alias);
+}
+
+function findMatchingNvmVersion(entries: string[], resolvedAlias: string): string | undefined {
+  if (isNvmBuiltInLatestAlias(resolvedAlias)) {
+    return entries[0];
+  }
+
+  const version = resolvedAlias.replace(/^v/, '');
+  return entries.find(entry => {
+    const entryVersion = entry.slice(1); // strip 'v'
+    return entryVersion === version || entryVersion.startsWith(version + '.');
+  });
+}
+
 function resolveNvmAlias(nvmDir: string, alias: string, depth = 0): string | null {
   if (depth > 5) return null;
 
   // If it looks like a version already (e.g., "v22.18.0" or "22"), return it
   if (/^\d/.test(alias) || alias.startsWith('v')) return alias;
+  if (isNvmBuiltInLatestAlias(alias)) return alias;
 
   try {
     const aliasFile = path.join(nvmDir, 'alias', ...alias.split('/'));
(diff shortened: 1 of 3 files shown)


YishenTu:
[P1] default -> node nvm aliases still fail

`resolveNvmAlias()` only follows file-backed aliases under `~/.nvm/alias/...`, but nvm also supports built-in aliases such as `node` (for example, `nvm alias default node`). In that common setup this code returns `null`, so the new GUI fallback still cannot find the nvm bin directory.

I verified that behavior against the PR code by simulating `alias/default = node`; the added tests only cover numeric and LTS aliases, so this gap would ship untested.

Michael-OvO:
Addressed. The failure occurred because **`node` is a built-in nvm alias**. The previous code incorrectly tried to resolve it as a file-based alias located under `~/.nvm/alias/...`, which caused the resolver to return `null`.

The resolver has been updated to treat built-in aliases such as `node` and `stable` as **terminal aliases** rather than file-backed ones. When resolving the default `nvm` binary, these aliases are now mapped to the highest installed `v...` Node version.

Regression tests were added to cover both the GUI PATH fallback and direct CLI discovery scenarios.

PR #391, opened 2026-03-22
fix(selection): stop clearing editor selection when sidebar gets focus

### LINKED ISSUE #344: Editor text selection not shown in chat context (regression in v1.3.68)
## Bug Description

Editor text selection stopped being included as context in the Claudian chat panel after upgrading from v1.3.67 to v1.3.68. Highlighting text in a note no longer shows the selection indicator ("X lines selected") and the `<editor_selection>` context is not attached to messages.

**Works in:** v1.3.67
**Broken in:** v1.3.68

## Steps to Reproduce

1. Install Claudian v1.3.68 (via BRAT)
2. Open any markdown note in **editing mode** (Live Preview or Source)
3. Select/highlight text in the editor
4. Observe the Claudian chat panel — no selection indicator appears
5. Send a message — no `<editor_selection>` context is included

## Expected Behavior

Selected text should appear as a context chip in the chat input area (e.g. "3 lines selected"), and be included as \`<editor_selection>\` in the message sent to Claude. This works correctly in v1.3.67.

## Suspected Cause

PR #318 (\`fix(browser-selection): improve security and view type matching\`) shipped in v1.3.68 refactored \`isBrowserLikeView()\` and consolidated \`BrowserSelectionController\` logic. The view-type matching change may be interfering with the routing between \`SelectionController\` (editor) and \`BrowserSelectionController\`, causing one to suppress or preempt the other.

## Workaround

Downgrade to v1.3.67 via BRAT ("Choose a version to install for a beta plugin").

## Environment

- **OS:** Fedora 43 (kernel 6.18.12-200.fc43.x86_64)
- **Obsidian:** Desktop (non-Flatpak)
- **Claudian:** v1.3.68 (broken), v1.3.67 (working)
- **Editor mode:** Live Preview (\`livePreview: true\`)

## Already Tried

- Uninstalling and reinstalling the plugin via BRAT
- Confirming editor is in Edit mode (not Reading mode)
- Full Obsidian restart

GeraldScott:
## Summary

- Fixes #344 (editor text selection still disappearing after #356)
- Removes `clearWhenMarkdownIsNotActive()` which aggressively cleared stored selection whenever `getActiveViewOfType(MarkdownView)` returned `null` — this happens when the Claudian sidebar receives focus (e.g., from indicator DOM updates), causing the selection indicator to flash and immediately disappear
- Reverts to v1.3.67 behavior: skip polling when no MarkdownView is active but preserve the stored selection; it clears naturally via `handleDeselection()` when the user returns to the editor and deselects

## Test plan

- [x] Open a markdown note in Live Preview mode, select text — indicator persists in chat panel
- [x] Send a message with selection — `<editor_selection>` context is included
- [x] Switch to a non-markdown view (e.g., settings) and back — selection preserved
- [x] Deselect text in editor — indicator clears
- [x] Unit tests pass (`SelectionController.test.ts`)

### DIFF
--- a/src/features/chat/controllers/SelectionController.ts
+++ b/src/features/chat/controllers/SelectionController.ts
@@ -6,21 +6,21 @@ import { type EditorSelectionContext, getEditorView } from '../../../utils/edito
 import type { StoredSelection } from '../state/types';
 import { updateContextRowHasContent } from './contextRowVisibility';
 
-/** Polling interval for editor selection (ms). */
 const SELECTION_POLL_INTERVAL = 250;
-/** Grace period for editor blur when handing focus to chat input (ms). */
 const INPUT_HANDOFF_GRACE_MS = 1500;
+const HIGHLIGHT_KEY = 'claudian-selection';
 
 export class SelectionController {
   private app: App;
   private indicatorEl: HTMLElement;
   private inputEl: HTMLElement;
+  private focusScopeEl: HTMLElement;
   private contextRowEl: HTMLElement;
   private onVisibilityChange: (() => void) | null;
   private storedSelection: StoredSelection | null = null;
   private inputHandoffGraceUntil: number | null = null;
   private pollInterval: ReturnType<typeof setInterval> | null = null;
-  private readonly inputPointerDownHandler = () => {
+  private readonly focusScopePointerDownHandler = () => {
     if (!this.storedSelection) return;
     this.inputHandoffGraceUntil = Date.now() + INPUT_HANDOFF_GRACE_MS;
   };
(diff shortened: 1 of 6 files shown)


YishenTu:
Added commit `ebd3eaec` with the following changes on top of the original fix:

**Broadened focus scope** — Replaced the `focus` listener on `inputEl` with a `focusin` listener on `contentEl`. `focusin` bubbles, so it fires when *any* element inside the sidebar gets focus (input, message area, buttons), not just the chat input.

**Preview mode highlight** — Added CSS Custom Highlight API (`::highlight(claudian-selection)`) fallback for preview/reading mode. When the native DOM selection is lost (e.g., input receives focus), stored `domRanges` are used to create a mock highlight. Native selection is preferred when still visible.

**Native-first highlight strategy** — For both edit and preview modes, `showHighlight()` now checks whether native highlighting is still visible before applying a mock. Stale mocks are cleaned up when native selection becomes visible again, preventing duplicate highlights.

**Refined clear logic** — Replaced the blanket early-return (from the original PR) with `clearWhenMarkdownContextIsUnavailable()`, which preserves selection only while focus is within the chat sidebar (`isFocusWithinChatSidebar()`). Focus leaving both the editor *and* the sidebar now correctly clears stale selection context.

**Test updates** — Added/updated tests for `focusScopeEl`, `focusin` wiring, CSS Highlight API set/delete, DOM range tracking and comparison, disconnected range filtering, and sidebar-vs-external focus clearing.

PR #456, opened 2026-04-09
feat: add OpenCode provider integration via ACP protocol

zion991-cmd:


YishenTu:
i am adding support in branch: feautre/opencode-support, thx for the effort though.

PR #465, opened 2026-04-10
fix: correct text dedup in streaming with extended thinking

mvanhorn:
## Summary

When extended thinking is enabled and the SDK delivers text only in the final assistant message (no incremental `text_delta` events), the response text is not displayed. Users see only the "Thought for Xs" block with an empty text area.

## Root cause

`isStreamTextEvent()` in `ClaudeChatRuntime.ts:1418` returned `true` for `content_block_start` events with `content_block.type === 'text'`. This set `sawStreamText = true` even though no actual `text_delta` was received. The dedup logic at line 1501 then skipped the only copy of the text from the assistant message.

The `content_block_start` event only announces that a text block will exist - it does not confirm incremental text delivery followed. Some SDK configurations (particularly with extended thinking) skip `text_delta` events entirely and put the full text in the final assistant message.

## Changes

Removed the `content_block_start` branch from `isStreamTextEvent()` so it only returns `true` for `content_block_delta` with `delta.type === 'text_delta'`. This ensures dedup only activates when text was actually delivered incrementally.

Updated 3 tests to match the new behavior:
- `content_block_start` with text type now correctly returns `false`
- New test verifies `content_block_delta` with `text_delta` still triggers dedup
- Auto-turn dedup reset test uses `text_delta` events

## Testing

All 5056 tests pass. Lint and typecheck clean.

Fixes #420

This contribution was developed with AI assistance (Codex).

PR #466, opened 2026-04-10
fix: add thinking content dedup to prevent duplicate thinking blocks

### LINKED ISSUE #406: thinking 过程显示重复内容
大家有没有遇到啊？
我用的qwen coding plan，qwen, glm, minimax, kimi模型都试了
展开thinking, 看到内容重复（至少2-3次）
但是，我在cc cli 以及 vs ide里用qwen coding plan都没有问题，所以过来问问。

mvanhorn:
## Summary

Thinking blocks are displayed 2-3 times when using third-party models with extended thinking. The SDK delivers thinking content via both stream events and the final assistant message. Text already had dedup via `sawStreamText`, but thinking had no equivalent.

## Root cause

`transformClaudeMessage.ts` yields `{ type: 'thinking' }` from both `content_block_delta` stream events (lines 258-259) and from the assistant message content blocks (lines 156-158). Text avoids this duplication through `sawStreamText` tracking in `ClaudeChatRuntime.ts`, but no equivalent `sawStreamThinking` tracking existed for thinking blocks.

## Changes

Added `sawStreamThinking` tracking parallel to `sawStreamText`:

- `types.ts`: Extended `ResponseHandler` interface with `sawStreamThinking`, `markStreamThinkingSeen()`, `resetStreamThinking()`
- `ClaudeChatRuntime.ts`: Added `isStreamThinkingEvent()` that detects `content_block_delta` with `thinking_delta`
- `ClaudeChatRuntime.ts`: Wired thinking dedup into both `routeMessage()` (persistent query path) and `queryViaSDK()` (cold start path)
- Added 4 unit tests for `isStreamThinkingEvent()`

## Testing

All 5059 tests pass. Lint and typecheck clean.

Fixes #406

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/src/providers/claude/runtime/ClaudeChatRuntime.ts
+++ b/src/providers/claude/runtime/ClaudeChatRuntime.ts
@@ -170,6 +170,7 @@ export class ClaudianService implements ChatRuntime {
   // Auto-triggered turn handling (e.g., task-notification delivery by the SDK)
   private _autoTurnBuffer: StreamChunk[] = [];
   private _autoTurnSawStreamText = false;
+  private _autoTurnSawStreamThinking = false;
   private _autoTurnCallback: ((result: AutoTurnResult) => void) | null = null;
   private turnMetadata: ChatTurnMetadata = {};
   private bufferedUsageChunk: StreamChunk & { type: 'usage' } | null = null;
@@ -584,6 +585,7 @@ export class ClaudianService implements ChatRuntime {
     this.currentConfig = null;
     this._autoTurnBuffer = [];
     this._autoTurnSawStreamText = false;
+    this._autoTurnSawStreamThinking = false;
     if (!preserveHandlers) {
       this.responseHandlers = [];
       this.currentAllowedTools = null;
@@ -816,6 +818,13 @@ export class ClaudianService implements ChatRuntime {
         this._autoTurnSawStreamText = true;
       }
     }
+    if (this.isStreamThinkingEvent(message)) {
+      if (handler) {
+        handler.markStreamThinkingSeen();
+      } else {
+        this._autoTurnSawStreamThinking = true;
+      }
+    }
 
     // Transform SDK message to StreamChunks
     for (const event of transformSDKMessage(message, this.getTransformOptions())) {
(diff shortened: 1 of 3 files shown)


YishenTu (commented):
I found two correctness issues in the thinking dedup change.

1. `isStreamThinkingEvent()` marks `sawStreamThinking` for any `thinking_delta`, but `transformSDKMessage()` only emits thinking chunks when `parentToolUseId === null`. That means a subagent `stream_event` can set the dedup flag without producing any visible thinking chunk, and the later main assistant thinking then gets skipped. This is a regression for multi-agent turns.

2. The fix still misses thinking delivered in `content_block_start`. `transformSDKMessage()` emits thinking from both `content_block_start` and `thinking_delta`, but `isStreamThinkingEvent()` only treats `thinking_delta` as streamed thinking. If a response carries reasoning entirely in the start block, the final assistant thinking block will still duplicate. The new unit test currently bakes in that mismatch by asserting `content_block_start` should return false.

I also ran the targeted Claude unit suites in the PR worktree; they passed, so this looks like a logic/coverage gap rather than an existing test failure.

mvanhorn:
Thanks, both clear calls. Pushed fa13be0.

1. `isStreamThinkingEvent()` now returns false when `message.parent_tool_use_id` is non-null. Subagent stream_events no longer arm the main-turn dedup flag, so a main-agent thinking block that follows a subagent turn renders normally instead of being skipped.

2. It also returns true when a `content_block_start` carries a non-empty `content_block.thinking`, matching the emission path in `transformSDKMessage` (parentToolUseId === null + truthy thinking). A reasoning-only response delivered entirely in the start block now dedups instead of rendering twice.

Tests: split the old 'content_block_start with thinking type always returns false' case into no-payload, empty, and non-empty, and added two subagent negative cases (thinking_delta and content_block_start). Same "keep dedup aware" direction as #465.

YishenTu:
Closing in favor of #510. That PR consolidates the Claude stream dedup fix at the normalized chunk boundary, so the thinking case from #466 is covered together with the related text path instead of maintaining parallel raw-event heuristics.

PR #464, opened 2026-04-10
fix: handle missing codexHome in Codex app-server initialize response

CodingOnStar:
#463 

fix: handle missing codexHome in Codex app-server initialize response

Make codexHome optional in the initialize payload and fall back safely
when deriving Codex runtime paths, preventing replace() crashes in Obsidian.

PR #469, opened 2026-04-10
fix: handle missing codexHome in Codex runtime context

Lebenswille:
## Summary

Handle missing `initialize.codexHome` in the Codex runtime context setup.

## Why

Some Codex CLI setups appear to return an initialize payload without `codexHome`. Claudian currently assumes that field is always present and crashes during path normalization with:

`Cannot read properties of undefined (reading 'replace')`

## What changed

- Made `normalizeTargetPath()` resilient to non-string input.
- Made `joinTargetPath()` skip null/empty path segments before normalizing.
- Added a fallback to `~/.codex` when `initializeResult.codexHome` is missing or blank.
- Added a unit test that covers the missing-`codexHome` path.

## Verification

Ran:

`npm test -- --runTestsByPath tests/unit/providers/codex/runtime/CodexRuntimeContext.test.ts`

Result:

- 1 test suite passed
- 3 tests passed

## User-visible impact

Claudian no longer crashes with a generic TypeError during Codex startup when `codexHome` is absent from the initialize payload.

### DIFF
--- a/src/providers/codex/runtime/CodexRuntimeContext.ts
+++ b/src/providers/codex/runtime/CodexRuntimeContext.ts
@@ -14,15 +14,19 @@ export interface CodexRuntimeContext {
 }
 
 function normalizeTargetPath(launchSpec: CodexLaunchSpec, value: string): string {
+  const rawValue = typeof value === 'string' ? value : String(value ?? '');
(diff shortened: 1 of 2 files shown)


YishenTu:
fixed in #464

PR #473, opened 2026-04-11
fix: skip syncConversationState during active streaming in switchToTab

### LINKED ISSUE #437: [Bug] Tab switching interrupts active streaming/agent in other tabs
## Environment
- **OS**: macOS (Darwin 25.3.0)
- **Claudian**: v1.3.72
- **Claude Agent SDK**: ^0.2.76
- **Model**: Claude Opus 4.6 (1M context)

## Description

When switching between Claudian tabs while an agent is actively streaming/working in one tab, the active conversation gets interrupted. The agent stops mid-task, and when switching back, the user must manually send a message (e.g., "继续") to prompt it to continue — at which point the agent has lost its working context and has to re-orient from scratch.

## Steps to Reproduce

1. Open Claudian in Obsidian
2. In **Tab A**, start a complex task (e.g., a codebase seminar that spawns subagents)
3. While Tab A's agent is actively working (streaming, running tool calls, etc.), switch to **Tab B**
4. Switch back to **Tab A**

## Expected Behavior

Each tab's conversation should run independently. Switching to Tab B should not affect Tab A's active agent/streaming.

## Actual Behavior

- Tab A's agent is interrupted upon switching away
- The UI shows `(background task completed)` — the auto-turn callback fires with tool activity but no text (`renderAutoTriggeredTurn` at Tab.ts:1224)
- The agent does **not** resume on its own — the user must manually send a message to prompt it to continue
- After receiving the manual prompt, the agent says "Let me orient myself and find where we left off" — it has lost its working context

## Code Analysis

I traced through the TypeScript source (`src/`) to identify the root cause. Here's what I found and what I ruled out.

### What I Ruled Out: `switchToTab → setSessionId → closePersistentQuery`

My initial hypothesis was that `switchToTab()` (TabManager.ts:148) calls `tab.service.setSessionId()` (line 192), which triggers `closePersistentQuery('session switch')` (ClaudianService.ts:1481). However, after deeper investigation, **this should only affect the target tab's own service**, because:

- Each tab creates its own `ClaudianService` instance (`new ClaudianService()` at Tab.ts:252)
- Each service spawns its own CLI process via `agentQuery()` → `createCustomSpawnFunction()` → `child_process.spawn()` (customSpawn.ts:33)
- Each service has its own `queryAbortController`, `persistentQuery`, and `messageChannel`
- `switchToTab()` only calls `setSessionId()` on the **target tab's** service, never the source tab's

So Tab B's session sync should be completely isolated from Tab A's active streaming.

### The `(background task completed)` Symptom

This message comes from `renderAutoTriggeredTurn()` (Tab.ts:1203-1238). It fires when:

1. A `turnComplete` message arrives in the response consumer loop (ClaudianService.ts:727)
2. No handler is registered (`handler` is null at line 656)
3. Buffered chunks have tool activity but no text content (Tab.ts:1222-1224)

This means **Tab A's handler was already unregistered** when the SDK delivered the turn completion. The handler's `onDone()` was called prematurely, causing `state.done = true` (line 1018), which broke the `while (!state.done)` yield loop (line 1057). Subsequent chunks from the CLI were routed to the auto-turn buffer instead.

### Possible Root Causes (Not Yet Confirmed)

Since the code-level isolation appears correct, the actual root cause may be in a layer I couldn't fully inspect:

**1. Claude Agent SDK internal shared state**
The `agentQuery()` function from `@anthropic-ai/claude-agent-sdk` may use shared resources (session files, lock files, IPC channels) between multiple `Query` instances. When Tab B's service calls `persistentQuery.interrupt()`, it may affect Tab A's process through a shared session directory or similar mechanism.

**2. Electron renderer throttling**
When the Claudian view becomes non-foreground (switching Obsidian leaves, not Claudian tabs), Electron/Chromium may throttle the renderer process. If the response consumer loop's `for await` (line 564) depends on timely event loop processing, throttling could cause the stdio pipe to stall or the CLI process to timeout waiting for input.

**3. Race condition in handler lifecycle**
If `isTurnCompleteMessage()` fires at just the right moment during a tab switch, the handler could be unregistered before all chunks are delivered. The handler lifecycle assumes single-turn processing (line 646-648: "Only one handler exists at a time because MessageChannel enforces single-turn processing"), but subagent completion may create edge cases.

**4. AbortSignal propagation**
`customSpawn.ts:44` registers `signal.addEventListener('abort', () => child.kill())`. If the AbortSignal from Tab B's `queryAbortController` somehow affects Tab A's child process (e.g., through Electron's cross-realm AbortSignal issues noted at customSpawn.ts:30-32), it could kill Tab A's CLI process.

### Suggested Debugging Steps

1. Add logging to `closePersistentQuery()` to confirm which service instance is being closed:
   ```typescript
   closePersistentQuery(reason?: string) {
     console.log(`[closePQ] reason=${reason}, handlerCount=${this.responseHandlers.length}, consumerRunning=${this.responseConsumerRunning}`);
   }
   ```

2. Add logging to `routeMessage()` when a turn completes with no handler:
   ```typescript
   if (isTurnCompleteMessage(message)) {
     console.log(`[turnComplete] hasHandler=${!!handler}, bufferLen=${this._autoTurnBuffer.length}`);
   }
   ```

3. Log when `onDone()` is called on a handler to find the premature termination:
   ```typescript
   onDone: () => {
     console.log(`[handler.onDone] id=${handlerId}`, new Error().stack);
   }
   ```

4. Check if multiple tab services' CLI processes share a PID group or session directory.

### Key Files

| Component | File | Lines |
|-----------|------|-------|
| `switchToTab()` | `src/features/chat/tabs/TabManager.ts` | 148-203 |
| `setSessionId()` | `src/core/agent/ClaudianService.ts` | 1475-1495 |
| `closePersistentQuery()` | `src/core/agent/ClaudianService.ts` | 380-433 |
| Response consumer loop | `src/core/agent/ClaudianService.ts` | 550-632 |
| Handler yield loop | `src/core/agent/ClaudianService.ts` | 1057-1068 |
| `renderAutoTriggeredTurn()` | `src/features/chat/tabs/Tab.ts` | 1203-1238 |
| Custom spawn (abort → kill) | `src/core/agent/customSpawn.ts` | 40-46 |
| `deactivateTab()` / `activateTab()` | `src/features/chat/tabs/Tab.ts` | 1043-1060 |

## Related Issues

- #396 — Similar interruption behavior during Teams/Agent tasks (different trigger, potentially same root cause)
- #417 — AbortSignal EventTarget crash (may be related to cross-realm signal issues in customSpawn.ts)

## Note

This is particularly disruptive for long-running agent tasks (codebase exploration, multi-file operations) that can take 5-10+ minutes. Users naturally want to use other tabs while waiting, but doing so kills the running task.

### LINKED ISSUE #449: When Claude is in the "Thinking..." phase (extended reasoning, before any response text appears) and I switch to another tab, the generation stops completely — 100% of the time.
Describe the bug

When Claude is in the "Thinking..." phase (extended reasoning, before any response text appears) and I switch to another tab, the generation stops completely — 100% of the time. There is no way to resume it; the response is simply lost.

Expected behavior

Each Claudian tab should work as an independent session, similar to opening two separate terminal windows. Switching to another tab should not interrupt or cancel an ongoing response in the background tab — regardless of whether Claude is still thinking or already writing.

Actual behavior

Switching away from a tab during the "Thinking..." phase always cancels the request. The response never appears when switching back to that tab.

Important distinction that helps narrow down the bug

Switch tabs during "Thinking..." (before any text appears) → generation stops, 100% reproducible
Wait until Claude has already started writing the response (text is visible), then switch tabs → the response continues and completes normally
This means the issue is not with the streaming response itself — text generation runs in the background correctly. The problem is specifically with how the plugin handles the thinking/reasoning phase. The two phases appear to be handled differently under the hood.

Steps to reproduce

Open two Claudian tabs
Send a prompt in Tab 1 that triggers extended thinking (the "Thinking..." indicator appears)
While still in the "Thinking..." phase — before any response text appears — switch to Tab 2
Switch back to Tab 1
Result: the response is gone, generation did not continue
Why this matters

The tab UI strongly implies that each tab is an isolated, independent session. Users naturally expect to run multiple tasks in parallel across tabs — the same way developers use multiple terminal windows. Currently, switching tabs during the thinking phase kills the request entirely, making parallel work impossible in practice.

Environment

Plugin: Claudian (Obsidian community plugin)

ChouChouHu:
## Summary

Fixes #437, fixes #449

When switching back to a tab that is actively streaming, `switchToTab()` unconditionally calls `syncConversationState()`, which calls `setSessionId()`. If the resolved session ID differs from the current one (e.g. after a forced restart from `applyDynamicUpdates`), it calls `closePersistentQuery("session switch")` — killing the CLI child process and terminating the active stream.

This adds an `isStreaming` guard so that tabs with active streams are left untouched during tab switches. The sync runs normally for idle tabs.

## Debug Evidence

Added `console.warn` to `closePersistentQuery`, `handler.onDone/onError`, `cancel()`, and `switchToTab` in the bundled `main.js`. Reproduction: start a streaming conversation in Tab A → switch to Tab B → switch back to Tab A.

```
switchToTab { from: "tab-A", to: "tab-B" }
closePersistentQuery { reason: "forced restart", handlerCount: 0 }
responseConsumer: iterator exhausted (normal end)

switchToTab { from: "tab-B", to: "tab-A" }
closePersistentQuery { reason: "session switch", handlerCount: 1 }  ← kills active stream
handler.onDone()
responseConsumer: iterator exhausted (normal end)
```

Call stack for the killing call:
```
ClaudianService.closePersistentQuery
  ← ClaudianService.setSessionId
  ← ClaudianService.syncConversationState
  ← TabManager.switchToTab
  ← ClaudianView.handleTabClick
```

## Change

```diff
// TabManager.ts — switchToTab()
 } else if (tab.conversationId && tab.state.messages.length > 0 && tab.service) {
+  if (!tab.state.isStreaming) {
     const conversation = this.plugin.getConversationSync(tab.conversationId);
     ...
     tab.service.syncConversationState(conversation, externalContextPaths);
+  }
 }
```

## Test plan

- [ ] Start a streaming conversation in Tab A
- [ ] Switch to Tab B while Tab A is streaming
- [ ] Switch back to Tab A — streaming should continue uninterrupted
- [ ] Verify idle tab switching still syncs conversation state correctly
- [ ] Verify `(background task completed)` no longer appears after tab switch

Tested on v2.0.2 by patching the bundled `main.js` directly — confirmed the fix resolves both #437 and #449.

### DIFF
--- a/src/features/chat/tabs/TabManager.ts
+++ b/src/features/chat/tabs/TabManager.ts
@@ -226,15 +226,20 @@ export class TabManager implements TabManagerInterface {
       if (tab.conversationId && tab.state.messages.length === 0) {
         await tab.controllers.conversationController?.switchTo(tab.conversationId);
       } else if (tab.conversationId && tab.state.messages.length > 0 && tab.service) {
-        // Tab already has messages loaded and runtime exists — passive sync only
-        const conversation = this.plugin.getConversationSync(tab.conversationId);
-        if (conversation) {
-          const hasMessages = conversation.messages.length > 0;
-          const externalContextPaths = hasMessages
-            ? conversation.externalContextPaths || []
-            : (this.plugin.settings.persistentExternalContextPaths || []);
+        // Tab already has messages loaded and runtime exists — passive sync only.
+        // Skip sync if the tab is actively streaming to avoid killing the CLI process.
+        // syncConversationState → setSessionId → closePersistentQuery("session switch")
+        // would terminate the running persistent query when session IDs diverge.
+        if (!tab.state.isStreaming) {
+          const conversation = this.plugin.getConversationSync(tab.conversationId);
+          if (conversation) {
+            const hasMessages = conversation.messages.length > 0;
+            const externalContextPaths = hasMessages
+              ? conversation.externalContextPaths || []
+              : (this.plugin.settings.persistentExternalContextPaths || []);
 
-          tab.service.syncConversationState(conversation, externalContextPaths);
+            tab.service.syncConversationState(conversation, externalContextPaths);
+          }
         }
       } else if (!tab.conversationId && tab.state.messages.length === 0) {
         // New tab with no conversation - initialize welcome greeting


YishenTu:
i already fixed this issue in #491, please tested it.

PR #484, opened 2026-04-13
feat: add Cursor Agent as third chat provider

darmawan01:
## Summary
Adds **Cursor Agent** as a third chat provider.

## Key changes
- Register Cursor provider via subprocess Cursor CLI (`agent -p`, `stream-json`)
- Session resume via `--resume`; workspace keyed by vault path (md5)
- Hydrate history from `~/.cursor/chats` SQLite store when `node:sqlite` is available
- Adds settings tab, env reconciler, and title/refine/inline auxiliary runners
- Model presets aligned with current Cursor CLI (`auto`, `composer-2`, etc.)
- Maps legacy `composer-1` to `auto` for CLI compatibility
- Fix unit test to avoid developer-specific absolute path + MD5 assertion
- Ignore local Cursor IDE `.cursor/` directory

## Screenshots
_TBD — will add screenshot in a follow-up._

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test`
- [x] `npm run build`
- [x] Manual smoke: select Cursor provider → send message → verify streaming + cancel/resume + history load

## Notes
- History hydration requires SQLite support (`node:sqlite`) to be available; otherwise Cursor history load is skipped gracefully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ yarn-error.log*
 # IDE
 .idea/
 .vscode/
+.cursor/
 *.swp
 *.swo
 *~
--- a/esbuild.config.mjs
+++ b/esbuild.config.mjs
@@ -94,6 +94,7 @@ const context = await esbuild.context({
     '@lezer/lr',
     ...builtins,
     ...builtins.map(m => `node:${m}`),
+    'node:sqlite',
   ],
   format: 'cjs',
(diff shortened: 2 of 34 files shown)


darmawan01:
image


image

snoopy-rui:
up

Issue #488, opened 2026-04-14
feat: "New Tab" should auto-open Claudian view if not already open


wc222:
### Problem

Currently, the **"Claudian: New tab"** command only works when the Claudian chat view is already open. If the view is closed, the hotkey does nothing — the user must first invoke **"Claudian: Open chat view"** manually before "New tab" can function.

### Expected Behavior

When the user triggers **"Claudian: New tab"**, if no Claudian view is currently open, the plugin should automatically open the Claudian view and then create a new tab in one step.

### Current Workaround

Press "Open chat view" (`⌘⌃A`) first, then press "New tab" — requiring two actions for what should be one.

---

### 问题

当前 **"Claudian: New tab"** 命令仅在 Claudian 聊天视图已打开时才生效。若视图未打开，快捷键无任何响应，用户必须先手动执行 **"Claudian: Open chat view"** 才能使用 New tab。

### 期望行为

当用户触发 **"Claudian: New tab"** 时，若 Claudian 视图尚未打开，插件应自动打开视图并创建新标签页，一步完成。

### 当前临时方案

先按 "Open chat view"（`⌘⌃A`），再按 "New tab"——本应一步完成的操作需要两步。

PR #507, opened 2026-04-16
Fix New tab when Claudian view is closed

YishenTu:
When `Claudian: New tab` runs with the chat view closed, the plugin now opens the view first and preserves the expected one-step behavior.

The plugin also tracks the last persisted tab layout so reopening from a closed state avoids creating a duplicate blank tab on a cold open while still adding a tab when a prior layout exists.

This branch routes tab-manager persistence through the plugin for that availability check and adds integration coverage for the closed-view, restored-layout, and max-tab-limit scenarios.

Testing: `npm run typecheck`, `npm run lint`, `npm run test`, and `npm run build`.


YishenTu:
fixed #488

Issue #480, opened 2026-04-12
Feature: route fine-grained tool input through Claudian's existing tool-delta API


LinFeng1023:
### What problem does this solve?

Claudian already has a solid external tool rendering surface:

- `registerToolDeltaHandler()`
- `registerToolRenderHandler()`
- `tool_delta` / `tool_flush`
- workspace event `claudian:tool-delta`

That part is already in place and feels like the right architecture.

The remaining gap is granularity.

For tools that generate large structured payloads such as HTML, diagrams, previews, or other long JSON-like content, external renderers still cannot reliably render progressively unless Claudian forwards finer-grained tool input updates through that existing API.

Right now, the public hook exists, but the token-level path still appears incomplete.

---

### What already exists in Claudian

From the current source, Claudian already has:

- generic external tool delta handlers
- generic external tool render handlers
- `tool_delta` and `tool_flush`
- support for external renderers in both `streaming` and `stored` modes

This is great. I am **not** asking for a brand-new external hook family.

---

### What seems to be missing

Two things still look incomplete for true progressive external tool rendering:

1. `transformSDKMessage()` does not currently forward `stream_event.content_block_delta` when `delta.type === 'input_json_delta'`
2. Fine-grained tool streaming does not appear to be an explicit Claudian-owned behavior / documented path yet

So the remaining ask is not "please add tool delta hooks", but:

> **Please feed fine-grained tool input deltas into Claudian's existing tool-delta API.**

---

### Why this is useful for Claudian itself

This is not only about one downstream plugin.

It would improve Claudian's external rendering platform more generally:

- richer progressive previews for long-running or structured tools
- lower perceived latency for users
- better support for external renderers that want to show intermediate structured state
- a cleaner foundation for future custom tool UIs

Examples of tool outputs that benefit from this:

- generated HTML / UI
- diagrams
- large structured previews
- long JSON-like tool input
- any renderer that wants to progressively update before final flush

---

### Possible implementation direction

Keep the current public API shape centered on:

- `tool_delta`
- `tool_flush`
- `registerToolDeltaHandler()`
- `registerToolRenderHandler()`

and extend the internals so fine-grained tool input can flow through that same path.

A possible direction would be:

1. explicitly enable fine-grained tool streaming in Claudian-owned SDK setup
2. track active streamed tool-use blocks by content block index
3. route `input_json_delta` into incremental `tool_use` updates

One possible payload shape could be either:

- an optional `partialJson` field on the existing delta event, or
[...]

PR #508, opened 2026-04-16
Stream incremental Claude tool input updates

YishenTu:
## Summary
- stream Claude `input_json_delta` events into cumulative `tool_use` updates in both persistent and cold-start runtime paths
- add tracked tool-input state so partial JSON can be merged across stream events and cleared at the right boundaries
- merge repeated subagent tool IDs so progressive updates refresh existing rows instead of duplicating them

## Testing
- `npm run typecheck`
- `npm run lint`
- `npm run test`
- `npm run build`


YishenTu:
fix #480

PR #512, opened 2026-04-16
feat: add current time to system prompt context

mentizm:
## Summary
- Adds `getCurrentTime()` utility to `src/utils/date.ts` that returns the 
  current local time in en-US format (e.g. "02:23 PM") using the user's system clock
- Imports and adds `getCurrentTime()` to the Time Context section of the 
  system prompt in `src/core/prompt/mainAgent.ts`

## Why
The system prompt already provides the current date as context. Adding the 
current time allows the AI to be aware of time-sensitive situations within 
the same day and give more precise time-aware responses.

The time is read from the user's local system clock via `new Date()` — 
no external calls, no hardcoded timezone.

## Test
Open a new chat and ask "what time is it?" — it should respond with the 
current local time.

### DIFF
--- a/src/core/prompt/mainAgent.ts
+++ b/src/core/prompt/mainAgent.ts
@@ -1,4 +1,4 @@
-import { getTodayDate } from '../../utils/date';
+import { getTodayDate, getCurrentTime } from '../../utils/date';
 
 export interface SystemPromptSettings {
   mediaFolder?: string;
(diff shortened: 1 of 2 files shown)


YishenTu:
context engineering 101, dont add time in system prompt.

what you can do is in customer instructions section, ask Claudian to run `bash: date` to get date and time.

mentizm:
Thanks for the feedback – makes total sense. Hardcoding time in the system prompt is stale by design. I've added the `bash: date` approach to my vault's CLAUDE.md instead. Good learning!

Issue #520, opened 2026-04-17
Feature request: Custom model IDs in settings (not just env vars)


kvarnelis:
## Problem

The model picker only shows 5 built-in aliases (Haiku, Sonnet, Sonnet 1M, Opus, Opus 1M). To add custom model IDs like `claude-opus-4-6` or pinned older versions, the only current option is env vars — but setting any of `ANTHROPIC_MODEL`, `ANTHROPIC_DEFAULT_OPUS_MODEL`, etc. **replaces** the entire picker rather than appending.

This makes it impossible to have both the built-in aliases AND additional pinned model IDs.

## Proposed solution

Add a `customModels` settings field (newline-separated model IDs) that appends to the default picker options instead of replacing them.

## Implementation approach

I prototyped this by patching the bundled `main.js` directly. Here's the approach translated to what the source files would need:

### 1. Add `customModels: ""` to Claude provider settings

In the Claude provider settings defaults, add a new field:

```ts
customModels: ""
```

### 2. Add it to the settings getter

In `getClaudeProviderSettings`, add:

```ts
customModels: config.customModels ?? DEFAULT_CLAUDE_PROVIDER_SETTINGS.customModels
```

### 3. Modify `getModelOptions` to parse and append

After `const models = [...DEFAULT_CLAUDE_MODELS]`, add:

```ts
const customModelIds = (claudeSettings.customModels || "")
  .split(/\r?\n/)
  .map(s => s.trim())
  .filter(s => s.length > 0);

for (const modelId of customModelIds) {
  const label = modelId
    .replace(/-/g, " ")
    .replace(/\b\w/g, l => l.toUpperCase());
  models.push({ value: modelId, label, description: "Custom model" });
}
```

### 4. Add textarea UI in Claude settings tab

After the `enableSonnet1M` toggle, add:

```ts
new Setting(container)
  .setName("Custom models")
  .setDesc("Add model IDs to the picker (one per line). Use [1m] suffix for 1M context, e.g. claude-opus-4-6[1m]")
  .addTextArea((text) => {
    text
      .setPlaceholder("claude-opus-4-6\nclaude-opus-4-6[1m]\nclaude-opus-4-5-20251101")
      .setValue(claudeSettings.customModels || "")
      .onChange(async (value) => {
        updateClaudeProviderSettings(settingsBag, { customModels: value });
        await context.plugin.saveSettings();
        context.refreshModelSelectors();
      });
    text.inputEl.rows = 6;
[...]

PR #521, opened 2026-04-18
Add custom Claude model entries in settings

YishenTu:
Adds a Claude settings textarea for custom model IDs that append to the built-in picker while preserving the existing environment-variable override behavior. Extracts shared Claude model option and label helpers so env- and settings-defined models resolve consistently, including friendlier labels for versioned and 1M model IDs. Reconciles active and title-generation model selections when custom entries are removed, and adds focused tests for option parsing, settings UI, and provider coordination. Verification: npm run typecheck, npm run lint, npm run test, npm run build.

per #520 request


kvarnelis:
Thank you!

PR #526, opened 2026-04-18
feat: line-range @mention via Option+K and Shift+drop

wanzirong:
## Summary

- **Option+K shortcut**: When an editor has an active selection, pressing Option+K (Alt+K on Windows/Linux) inserts `@filename#startLine-endLine` into the chat input. At send time, the referenced lines are read from the vault and injected as `<editor_selection path="..." lines="...">` XML into the prompt.
- **Reading mode support**: When no line numbers are available (reading/preview mode), Option+K instead inlines the selected text directly as an `<editor_selection>` block.
- **Shift+drop**: Holding Shift while dragging a file from the Obsidian file list onto the Claudian panel inserts `@filepath` into the input (without opening the file).

## Technical Notes

- Uses `document` capture-phase listeners for Shift+drop to intercept before Obsidian's own drop handler.
- `e.code === 'KeyK'` (not `e.key`) for Option+K to work correctly on macOS where Option+K produces the composed character `˚`.
- Send-time resolution happens in `InputController.sendMessage` (async) before `runtime.prepareTurn`, keeping `prepareTurn` synchronous.
- Manually typed `@file#line` is never resolved — only shortcut-inserted mentions are whitelisted via `FileContextState.attachLineRangeMention`.

## Test Plan

- [ ] In source mode, select lines in editor → press Option+K → verify `@filename#start-end` inserted in input
- [ ] Send the message → verify `<editor_selection>` XML with correct line content appears in prompt
- [ ] In reading mode, select text → press Option+K → verify `<editor_selection>` block inlined directly
- [ ] Press Option+K with no selection (cursor only) → verify nothing inserted
- [ ] Shift+drag a file onto Claudian panel → verify `@filepath` inserted
- [ ] Normal drag (no Shift) → verify file opens as usual, no mention inserted
- [ ] Manually type `@file#1-5` → send → verify no XML injection


YishenTu:
current selection indicator already supports selected content as well as line number, and it's all automatic.

what's new in this pr?

wanzirong:
Thanks for the context! Here's what's new:

  Inspiration: This is modeled after the Claude Code VS Code extension's line-range mention feature (https://code.claude.com/docs/en/vs-code).

  What's different from the existing selection indicator:

  The existing selection indicator automatically captures the current active editor selection — which is great, but it's limited to a single selection at a time and is replaced whenever the selection changes.

  This PR adds:

  1. Option+K (Mac) / Alt+K (Windows/Linux) — Press the shortcut while text is selected to pin that selection as an @filename#startLine-endLine mention in the input. You can do this multiple times to reference multiple independent line ranges from the same or different files in a single message.
  2. Shift+drag a file onto the Claudian panel to insert @filepath as a mention token without opening the file.
  3. At send time, each pinned mention is resolved into <editor_selection path="..." lines="...">…content…</editor_selection> XML so the model sees the exact lines.

  The key difference: the existing indicator is ephemeral and single-selection. This lets you deliberately pin multiple ranges before hitting send.

YishenTu:
i like the idea about the multi selection insertion, but i think the file-drag feature is not aligned with current design, the chip is to "link" a session to a note, not to display all the files that has been mentioned.

if i were to support multi-line selection, i will simply create a new hotkey to insert the selected content path and line number, plain and simple.

PR #545, opened 2026-04-21
feat: register Switch to tab 1..9 commands

Andr3y-:
Exposes nine Obsidian commands (switch-to-tab-1 .. switch-to-tab-9) that activate the Nth tab in the current Claudian view. Surfacing them through addCommand (rather than a view-scoped DOM keydown listener) means:

  - Users rebind in Settings → Hotkeys like any other command.
  - checkCallback hides slots above the current tab count so the command palette stays tidy.
  - Fires from anywhere in Obsidian, not only when the Claudian view has DOM focus.

No default chord is assigned. ⌘/⌥/Ctrl + digit all collide with common OS / browser bindings and the right choice varies per user, so we leave it to the user via Hotkeys. A README note would suggest ⌥1..9 on macOS and Alt+1..9 elsewhere.

Tests: +5 cases in tests/integration/main.test.ts
  - registers all nine commands with the expected names
  - switch-to-tab-1 calls TabManager.switchToTab(tabs[0].id)
  - switch-to-tab-3 is unavailable when only 2 tabs exist
  - unavailable when no Claudian leaf is open
  - unavailable when leaf has no TabManager yet

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -168,6 +168,33 @@ export default class ClaudianPlugin extends Plugin {
       },
     });
 
+    // `Switch to tab N` commands for N = 1..9. Exposed through Obsidian's
+    // addCommand so users can rebind in Settings → Hotkeys. No default
+    // chord is assigned; typical binding is ⌘/⌥ + digit but that varies
+    // by OS and user preference. checkCallback hides commands for slots
+    // that exceed the current tab count, keeping the command palette
+    // tidy.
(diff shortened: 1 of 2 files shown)


zhenhuaw-me:
@YishenTu do you mind to merge this PR? Thanks! I think it's pretty helpful to improve the efficiency when there are multiple sessions.

Consider it's 3+ month aged PR, if you see any compatibility risk, I can help to verify if you prefer.

PR #554, opened 2026-04-24
Update Codex model defaults for GPT-5.5

YishenTu:
## Summary
- centralize built-in Codex model ids, labels, fast-tier copy, and runtime fallbacks in `src/providers/codex/types/models.ts`
- switch Codex runtime, settings, and UI defaults to the shared GPT-5.5 primary model instead of scattered hardcoded strings
- migrate persisted legacy `gpt-5.4` Codex selections and saved provider projections to `gpt-5.5` so upgraded users keep aligned UI, runtime, and fast-tier behavior

## Testing
- npm run typecheck
- npm run lint
- npm run test
- npm run build

### DIFF
--- a/src/core/providers/ProviderSettingsCoordinator.ts
+++ b/src/core/providers/ProviderSettingsCoordinator.ts
@@ -68,6 +68,17 @@ function normalizeReasoningValue(
   return uiConfig.getDefaultReasoningValue(model);
 }
 
+function normalizeProviderModel(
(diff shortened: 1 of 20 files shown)


onezibo:
Hello, could you please release a new version to support this update? Additionally, I noticed that the context window for GPT-5.5 is 400K (and only supports this), I hope the context window can also be updated accordingly.

PR #563, opened 2026-04-25
fix: apply permission mode changes live

MainHanzo:
## Summary

This fixes permission mode changes made from the chat toolbar so they take effect immediately in the active conversation instead of only affecting later runtime setup.

- Adds `updatePermissionMode(mode)` to the shared chat runtime interface.
- Routes toolbar permission-mode changes through the active tab runtime after saving settings and updating UI state.
- Implements Claude live permission updates with `persistentQuery.setPermissionMode(...)`, preserving Claudian's existing mode mapping (`yolo -> bypassPermissions`, `plan -> plan`, Safe/normal -> configured Claude safe mode).
- Adds a Codex runtime no-op implementation because Codex reads permission settings when starting or resuming a turn.
- Covers switching into YOLO, into plan, and back from YOLO to Safe during an active persistent Claude query.

## Tested

- `npx jest --selectProjects unit --runTestsByPath tests/unit/providers/claude/runtime/ClaudianService.test.ts tests/unit/features/chat/tabs/Tab.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- Manual Obsidian smoke test: switched permission mode mid-conversation and confirmed the updated mode took effect live.

### DIFF
--- a/src/core/runtime/ChatRuntime.ts
+++ b/src/core/runtime/ChatRuntime.ts
@@ -29,6 +29,7 @@ export interface ChatRuntime {
   ): void;
   reloadMcpServers(): Promise<void>;
   ensureReady(options?: ChatRuntimeEnsureReadyOptions): Promise<boolean>;
+  updatePermissionMode(mode: string): Promise<void>;
   query(
     turn: PreparedChatTurn,
     conversationHistory?: ChatMessage[],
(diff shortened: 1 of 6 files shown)


YishenTu:
it's already working without a restart.

PR #588, opened 2026-04-30
fix: include local settings source so .claude/settings.local.json is loaded by SDK

wanzirong:
## Summary

- `settingSources` was missing `'local'`, so the SDK never loaded `.claude/settings.local.json`
- Added `'local'` to both the SDK options array and the restart-detection key string in `buildPersistentQueryConfig`

## Test plan

- Add config to `.claude/settings.local.json` (e.g. `pluginConfigs`, `permissions.allow`)
- Start a conversation in Claudian — settings should now take effect

### DIFF
--- a/src/providers/claude/runtime/ClaudeQueryOptionsBuilder.ts
+++ b/src/providers/claude/runtime/ClaudeQueryOptionsBuilder.ts
@@ -120,7 +120,7 @@ export class QueryOptionsBuilder {
       mcpServersKey: '', // Dynamic via setMcpServers, not tracked for restart
       pluginsKey,
       externalContextPaths: externalContextPaths || [],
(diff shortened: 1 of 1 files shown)


YishenTu:
Thanks for the fix. I verified with the installed Claude SDK that `local` is the correct setting source for `.claude/settings.local.json`, but this PR only updates the main persistent runtime path and leaves cold-start auxiliary queries plus runtime command probing on `['project']` / `['user', 'project']`. It also leaves CI red because the QueryOptionsBuilder expectations were not updated.

I opened #594 with the same SDK behavior centralized in one Claude helper and reused across persistent chat, cold-start queries, and command probing, with tests updated across those call sites. Closing this one in favor of #594.

PR #622, opened 2026-05-10
Fix Codex provider refresh when enabling it in settings

cyruszh0u:
## Bug

When enabling the Codex provider in settings, the model switcher did not immediately reflect the provider availability clearly enough in the current UI flow.

Users often had to open a new chat tab before they could actually switch to a Codex model, which made it feel like Codex had not been enabled successfully.

## Fix

- refresh existing chat views after enabling Codex
- re-run provider availability handling for already open tabs before re-rendering model selectors
- add a regression test covering the refresh path when Codex is enabled


YishenTu:
enable/disable a provider will refresh the model selector instantly for new session, already started session is locked with the original provider, hence you wont see model selector update, this is not a bug.

PR #653, opened 2026-05-15
fix: stabilize composer expand and collapse behavior

### LINKED ISSUE #649: Feature request: Make input panel resizable with a top splitter handle
The height of the input field can be adjusted freely, allowing for maximum visibility when the field contains a lot of text.

nycanshu:
## Summary

  This PR fixes the composer input resizing issue (#649).

  The main problem was that long text in the composer could make the input area jump or resize in an awkward way. It was especially noticeable when
   pasting a large prompt, typing past the compact input height, or adding an image and then collapsing the composer.

  I changed the composer behavior to be more predictable:

  - Short prompts stay in the normal compact composer.
  - Long prompts auto-expand when the compact box overflows.
  - The expand/collapse button lets the user manually switch between the larger composer and the compact view.
  - If the user manually collapses a long prompt, it stays collapsed instead of expanding again immediately.
  - Clearing or sending the message resets the composer back to the normal compact state.
  - Image/file/context chips no longer break or clip the composer layout.

  I'm also attaching a short video to show the before/after behavior and the expected flow.

  Video: https://drive.google.com/file/d/1JZ6K14DEjvvwdANw1KzAP2qxjmItmYJn/view?usp=sharing

  ## What Changed

  - Added an explicit composer state model for `compact`, `expanded`, and `manual-collapsed`.
  - Replaced the older dynamic textarea resize logic with state-based behavior.
  - Added an expand/collapse button in the composer toolbar.
  - Updated the CSS so compact/collapsed layouts remain stable and long text scrolls inside the textarea.
  - Fixed layout handling when image/context chips are present.
  - Added tests for the composer state transitions, toolbar behavior, tab wiring, and CSS clipping regression.

  ## Issue Resolved

  Resolves #649

  ## Testing

  Tested locally in Obsidian on macOS with:

  - Short text input
  - Long typed input
  - Large pasted text
  - Manual expand/collapse
  - Clearing the input
  - Sending the message
  - Long text with an attached image
  - Collapsing after image attachment

  Also ran:

  ```bash
  npm run typecheck
  npm run lint
  npm run test
  npm run build
```


nycanshu:
Hii @YishenTu, looking for your valuable suggeston/review/comment on this PR. 👀

YishenTu:
After testing the PR in Obsidian, the main issue I see is that the transition from the normal composer to the extended composer feels too sudden. The jump is noticeable because long typed or pasted content immediately changes the composer from compact height to the large extended height.

Decision: separate normal auto-growth from extended mode.

Suggested behavior:

- Normal composer stays in the default mode while typing or pasting.
- In normal mode, the textarea grows one line at a time up to a capped normal max height.
- Once that normal max is reached, the textarea scrolls internally instead of switching to extended mode.
- Extended mode is entered only by clicking the toolbar icon.
- Clicking the icon toggles between normal and extended size.
- Send and clear should reset the composer back to normal mode.

This likely means we can simplify the state model. `manual-collapsed` is mainly needed because the current implementation auto-enters expanded mode. If auto-expand is removed, the model can probably become just normal/expanded, with normal height derived from textarea content and capped by CSS/measurement.

nycanshu:
> After testing the PR in Obsidian, the main issue I see is that the transition from the normal composer to the extended composer feels too sudden. The jump is noticeable because long typed or pasted content immediately changes the composer from compact height to the large extended height.
> 
> Decision: separate normal auto-growth from extended mode.
> 
> Suggested behavior:
> 
> * Normal composer stays in the default mode while typing or pasting.
> * In normal mode, the textarea grows one line at a time up to a capped normal max height.
> * Once that normal max is reached, the textarea scrolls internally instead of switching to extended mode.
> * Extended mode is entered only by clicking the toolbar icon.
> * Clicking the icon toggles between normal and extended size.
> * Send and clear should reset the composer back to normal mode.
> 
> This likely means we can simplify the state model. `manual-collapsed` is mainly needed because the current implementation auto-enters expanded mode. If auto-expand is removed, the model can probably become just normal/expanded, with normal height derived from textarea content and capped by CSS/measurement.

Okay, understood. I’ll update it so normal mode only auto-grows to a capped height with internal scroll, and expanded mode is only controlled by the toolbar toggle. I’ll also simplify the state handling based on your suggestions.

nycanshu:
Hi @YishenTu , I've rebased onto the latest main and addressed your review feedback:

  - Dropped manual-collapsed and auto-expand-on-overflow — now just compact / expanded, per your suggestion.
  - Compact mode is fully content-driven (grow + shrink line-by-line, no jiggle); expanded mode pins to a single shared cap.
  - Hybrid behavior: emptying text while expanded auto-collapses back to baseline.
  - Single source of truth for the cap — CSS custom properties on .claudian-input-wrapper, read by Tab.ts via getComputedStyle. Pixel-perfect
  parity between type-to-max and click-to-expand.
  - Also fixed a related bug: chip context row was uncapped in expanded mode, which could squeeze the textarea when many images were attached.

  This branch supersedes #662 — the two-step bounce fix lives inside resizeCompactTextarea, so the standalone textareaResize.ts module is no longer needed and is deleted in the rebase.

 All checks pass locally: typecheck, lint, 5296 unit tests, 205 integration tests, build. 

Thanks for the review!

YishenTu:
for the sake of discussion, i already fixed the flashy input resize issue. now when typing across the threshold, a stable new line height is inserted.

so i am thinking whether we still need to add an expand button just to maximize the input box.

nycanshu:
Hey @YishenTu , yeah you're right, the threshold jump is fixed on main, the expand button doesn't really add much. Internal scroll past the cap already covers the main use case, and a separate toggle just adds UI weight for something users may not even notice.

Happy to close this PR as implemented on your side. Thanks for taking the time to review and discuss, learned a lot going through this one 🙌

 You can close this pr.

Issue #614, opened 2026-05-08
Windows: spawn EINVAL when using OpenCode provider


pronzcw:
## Environment
- OS: Windows 10
- Claudian version: 2.0.11
- OpenCode installed via: npm install -g opencode-ai

## Problem
When trying to use OpenCode provider in Claudian on Windows, it fails with spawn errors:

1. **With default settings:** Error: spawn opencode ENOENT
2. **With full path to opencode.cmd:** Error: spawn EINVAL
3. **With full path to opencode.exe:** Error: spawn ENOENT

## Root Cause
Windows requires shell: true when spawning .cmd files. Current implementation likely uses shell: false.

## Suggested Fix
Use shell: true option on Windows when executing .cmd files, or spawn the .exe directly.

YishenTu:
put the opencode path into setting.

chenhui-su:
The same, "put the opencode path into settings" didn't help.  It seems that some environment variable passing is broken, or the plugin is making a mistake, for example, thinking that the opencode path needs to end with some extension.

YishenTu:
Resolved by #712. The fix covers Windows OpenCode npm installations by discovering .exe/.cmd shims, wrapping .cmd launches explicitly, and cleaning up the full process tree on shutdown.

PR #663, opened 2026-05-18
fix: support Windows ACP subprocess spawn in Opencode integration

McClone:
## Summary

Fix `Error: spawn EINVAL` on Windows when launching the OpenCode ACP subprocess.

## Changes

- `AcpSubprocess.ts`: Added `shell: true` on Windows for `.cmd` file support
- Replaced `SIGTERM`/`SIGKILL` (POSIX signals not supported on Windows) with `SIGINT` and `taskkill` fallback

### DIFF
--- a/src/providers/acp/AcpSubprocess.ts
+++ b/src/providers/acp/AcpSubprocess.ts
@@ -1,7 +1,8 @@
 import { type ChildProcessWithoutNullStreams, spawn } from 'node:child_process';
 import type { Readable, Writable } from 'node:stream';
 
-const SIGKILL_TIMEOUT_MS = 3_000;
+const IS_WIN32 = process.platform === 'win32';
+const KILL_TIMEOUT_MS = 3_000;
 const STDERR_BUFFER_LIMIT = 8_000;
 
 export interface AcpSubprocessLaunchSpec {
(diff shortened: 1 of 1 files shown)


McClone:
[https://github.com/YishenTu/claudian/issues/614](https://github.com/YishenTu/claudian/issues/614)

YishenTu (changes requested):
Findings:

- High: `src/providers/acp/AcpSubprocess.ts:50` uses `shell: true` on Windows while still passing `args`. The ACP callers pass `--cwd=<vault path>`, so common Windows vault paths with spaces or shell metacharacters can be parsed by `cmd.exe` instead of delivered as one argument. This can make OpenCode fail to start for paths like `C:\Users\Me\My Vault`, and it is also the class of issue Node warns about for shell execution. Prefer the existing Codex pattern: only wrap `.cmd` shims through `cmd.exe` and quote the command plus each argument explicitly.

- High: `src/providers/acp/AcpSubprocess.ts:106` schedules `taskkill /t` only after the timeout, but line 119 first kills the shell process created by `shell: true`. If that shell exits, cleanup clears the timer before `taskkill` runs, leaving the actual OpenCode process orphaned. Windows shutdown should either avoid the shell wrapper or use `taskkill /pid <pid> /t` as the primary termination path and wait for the process tree to exit.

Reviewed `origin/pr/663` against `origin/main`; `git diff --check` passed. I did not run the full test suite because these are static lifecycle/spawn issues.

LeonhardtHUST:
same subprocess spawn failure occurred. fix required

YishenTu:
Closing as superseded by #712. PR #712 resolves the Windows OpenCode ACP launch issue with explicit .exe/.cmd CLI discovery, targeted .cmd shim wrapping, and process-tree cleanup, without enabling shell: true for all Windows ACP launches.

PR #676, opened 2026-05-20
Support OpenCode ACP thinking levels

YishenTu:
Adds OpenCode ACP thought_level discovery and applies detached effort through the ACP effort config option. Persists selected-model thinking options while keeping the full OpenCode catalog in memory, including settings catalog loading and metadata warmup. Updates ACP, OpenCode runtime, settings, and UI tests.\n\nTests: npm run test; npm run typecheck && npm run lint -- --quiet; npm run build.


Easonnotsing:
Thank you very much

PR #704, opened 2026-05-29
[codex] Fix Command/Ctrl+Enter send shortcut

iblueer:
## Summary

Fix the optional Command/Ctrl+Enter send shortcut so macOS Command+Enter works reliably in the Obsidian chat view.

## Why

The existing textarea keydown handler checked `metaKey` on macOS, but Command+Enter can be intercepted before it reaches that handler in Obsidian/Electron. Ctrl+Enter could still work because it reached the textarea handler, which made the bug appear macOS-specific.

## Changes

- Keep the input keydown path accepting either Command or Ctrl when the send-shortcut setting is enabled.
- Add an Obsidian view `Scope` handler for `Meta+Enter`, limited to the active chat textarea and the existing setting.
- Add unit coverage for both the tab input handler and scoped Meta+Enter behavior.

## Validation

- `npm run test -- tests/unit/features/chat/ClaudianView.test.ts tests/unit/features/chat/tabs/Tab.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`

### DIFF
--- a/src/features/chat/ClaudianView.ts
+++ b/src/features/chat/ClaudianView.ts
@@ -614,6 +614,26 @@ export class ClaudianView extends ItemView {
     // View scopes are the Obsidian-owned boundary for main-area tab hotkeys.
     // Returning false consumes Escape before Obsidian uses it for pane navigation.
     this.scope = new Scope(this.app.scope);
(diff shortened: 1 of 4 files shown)


YishenTu:
fixed in #715

PR #705, opened 2026-05-30
fix(subagents): Stop hook false-positive on completed async subagents

songwunil:
## Summary

The Stop hook can block a session from ending forever, even when no subagents
are actually running. `SubagentManager.hasRunningSubagents()` counts the raw
size of `pendingAsyncSubagents` / `activeAsyncSubagents`:

```ts
return this.pendingAsyncSubagents.size > 0 || this.activeAsyncSubagents.size > 0;
```

Completed / errored / orphaned async subagents can linger in those maps:

- a completion notification (`handleAsyncSubagentResult`) arrives **before**
  `handleTaskToolResult` has registered the entry under its `agent_id`, so the
  notification is dropped and the entry is later inserted as `running` and never
  cleared; or
- a terminal `TaskOutput` result is otherwise not reconciled out.

When that happens, `hasRunningSubagents()` keeps returning `true` and the Stop
hook (`createStopSubagentHook` -> `STOP_BLOCK_REASON`) blocks **every** turn-end
with no satisfiable remedy: `TaskList` is empty and `TaskOutput task_id=... block=true`
reports the task as not found, yet the model is told to wait for it. The result
is an endless re-invocation loop that burns tokens (observed in the field while
the user was away from the session).

## Fix

Count by the actual `asyncStatus` instead of raw map size, and reconcile
terminal entries out on read. Only genuinely `pending` / `running` subagents
report as running, so the gate is **not** weakened for live subagents.

## Tests

Adds three `SubagentManager` cases:

- terminal entries lingering in both maps -> `hasRunningSubagents()` is `false`
  and the entries are reconciled out (maps do not leak);
- a genuinely `running` entry alongside a completed one -> still `true`, the live
  entry is kept and the completed one removed;
- a still-`pending` entry awaiting its `agent_id` -> still `true`.

## Verification

- `npm run typecheck` - clean
- `npm run lint` (changed files) - clean
- `SubagentManager.test.ts` - 100 passed (97 existing + 3 new)

## Notes

The reconciliation is defensive: it fixes the symptom (lingering terminal
entries blocking Stop) without changing the lifecycle transitions. If you would
prefer to additionally close the underlying registration race in
`handleAsyncSubagentResult` / `handleTaskToolResult`, happy to follow up.

### DIFF
--- a/src/features/chat/services/SubagentManager.ts
+++ b/src/features/chat/services/SubagentManager.ts
@@ -487,8 +487,35 @@ export class SubagentManager {
   // ============================================
 
   public hasRunningSubagents(): boolean {
-    // pendingAsyncSubagents: awaiting agent_id; activeAsyncSubagents: only holds running entries
-    return this.pendingAsyncSubagents.size > 0 || this.activeAsyncSubagents.size > 0;
+    // Count by actual asyncStatus instead of raw Map size. Completed/errored/
+    // orphaned async subagents can linger in these maps when a completion notice
+    // arrives before the agent_id is registered, or when a terminal TaskOutput
+    // result is dropped. Counting size alone then makes the Stop hook block every
+    // turn-end with no satisfiable remedy (TaskList is empty and TaskOutput
+    // reports the task as not found), forcing the model to re-invoke indefinitely.
+    // Reconcile terminal entries out here and only report "running" for genuinely
+    // pending/running subagents, so the gate is never weakened for live ones.
+    const isTerminal = (subagent: SubagentInfo | undefined): boolean =>
+      subagent?.asyncStatus === 'completed' ||
+      subagent?.asyncStatus === 'error' ||
+      subagent?.asyncStatus === 'orphaned';
+
+    let running = false;
+    for (const [taskToolId, subagent] of this.pendingAsyncSubagents) {
+      if (isTerminal(subagent)) {
+        this.pendingAsyncSubagents.delete(taskToolId);
+      } else {
+        running = true;
+      }
+    }
+    for (const [agentId, subagent] of this.activeAsyncSubagents) {
+      if (isTerminal(subagent)) {
+        this.activeAsyncSubagents.delete(agentId);
+      } else {
+        running = true;
+      }
+    }
+    return running;
   }
 
   // ============================================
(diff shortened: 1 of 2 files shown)


YishenTu:
## Code review findings

### Critical

None.

### High

#### 1. Early completion-notification race described by the PR is still not fixed

- Location: `src/features/chat/services/SubagentManager.ts:324-357`, `:428-435`, `:489-518`
- Evidence: `handleAsyncSubagentResult()` still drops a completion/error notification when the agent is not yet in `activeAsyncSubagents`:

```ts
const subagent = this.activeAsyncSubagents.get(agentId);
if (!subagent || subagent.asyncStatus !== 'running') {
  return undefined;
}
```

Then `handleTaskToolResult()` can later register that same agent as `asyncStatus = 'running'`, and `hasRunningSubagents()` treats non-terminal entries as running.

- Impact: The PR body's main race can still occur: completion arrives before `agent_id` registration, gets discarded, then the subagent is added as running indefinitely. The Stop hook can still block forever.
- Suggested fix: Buffer terminal `async_subagent_result` events by `agentId` when no active subagent exists, then apply the buffered terminal status in `handleTaskToolResult()` once the `agentId` is parsed. Add a public-method test in this order:
  1. create async task
  2. call `handleAsyncSubagentResult('agent-123', 'completed', ...)`
  3. call `handleTaskToolResult(...agent-123...)`
  4. assert `hasRunningSubagents()` is false

### Medium

#### 2. Reconciliation deletes primary map entries but leaves secondary indexes stale

- Location: `src/features/chat/services/SubagentManager.ts:70-73`, `:505-513`; related call site `src/features/chat/controllers/StreamController.ts:1132-1143`
- Evidence: The new reconciliation removes entries from `pendingAsyncSubagents` / `activeAsyncSubagents`, but does not clear `taskIdToAgentId` or `outputToolIdToAgentId`. `isLinkedAgentOutputTool()` still reads `outputToolIdToAgentId`, and `StreamController` treats `isLinked === true` as handled even when `handleAgentOutputToolResult()` returns `undefined`.
- Impact: A reconciled terminal active subagent can leave stale tool/task indexes behind, causing leaks and possible false-positive handling for later tool results with the same id.
- Suggested fix: Centralize async-subagent removal in a helper that deletes the active entry plus any `taskIdToAgentId` / `outputToolIdToAgentId` references for that agent, and use it from both terminal handlers and `hasRunningSubagents()` reconciliation.

### Low

None.

PR #727, opened 2026-06-02
feat: add model discovery via Anthropic /v1/models API

jesper0xyz:
- Add ClaudeModelDiscoveryService to fetch available models

- Add discoveredModels to ClaudeProviderSettings with persistence

- Add Discover/Clear buttons in settings UI

- Prefer discovered models over hardcoded defaults when available

- Improve non-Error throw messages in ClaudeChatRuntime

- Update CLAUDE.md with path aliases and command docs


jesper0xyz (commented):
viewed

PR #742, opened 2026-06-05
feat: add Remote Control support for the Claude provider

liqing-ustc:
## Summary

Adds **Remote Control** support to the Claude provider, so a Claudian-hosted session can be driven from [claude.ai/code](https://claude.ai/code) and the Claude mobile app, plus the fixes needed to make remotely sent prompts render in Obsidian.

## What's included

1. **Manual command** — "Enable remote control (current session)" calls the agent SDK's `enableRemoteControl` control request on the active session's runtime and surfaces the returned session URL via a `Notice`.

2. **"Remote control by default" toggle** — Claude provider settings → Experimental (default **off**). Auto-enables Remote Control on every session at init. Headless/SDK mode cannot reuse the CLI's persistent bridge, so each session (including after an Obsidian reload) registers a fresh claude.ai session; this is noted in the toggle description.

3. **Render remotely sent user prompts** — Remote Control turns arrive without a handler and the SDK stream does not echo the originating user message, so previously only the assistant reply appeared. The runtime now recovers the prompt from the on-disk session transcript (anchored on the assistant uuid, with a fallback to the last real user message to tolerate the transcript-write race) and passes it via `AutoTurnResult.precedingUserMessage`; the renderer shows it before the assistant reply, de-duplicating against messages already displayed.

## Implementation notes

- `enableRemoteControl` is added as an **optional** method on the `ChatRuntime` interface; only the Claude provider implements it (the underlying `@anthropic-ai/claude-agent-sdk` exposes it on `Query` but does not declare it in its `.d.ts`, so the implementation casts).
- Persistent/resumable Remote Control across a process restart is not available in headless/SDK mode today — see anthropics/claude-code#30447. The default-off toggle reflects the resulting "new session per reload" behavior.

## Testing

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run build` ✓
- `npm run test` — 240 suites / 5677 tests ✓
- Manually verified in Obsidian: manual command registers a session that appears on claude.ai/code; messages sent from the web render in Obsidian (both user prompt and assistant reply), including the transcript-write-race path.

### DIFF
--- a/src/core/runtime/ChatRuntime.ts
+++ b/src/core/runtime/ChatRuntime.ts
@@ -43,6 +43,14 @@ export interface ChatRuntime {
   isReady(): boolean;
   getSupportedCommands(): Promise<SlashCommand[]>;
   getAuxiliaryModel?(): string | null;
+
+  /**
+   * Registers (or unregisters) this session for remote control via the
+   * underlying agent SDK, so it can be driven from a paired Claude client.
+   * Only providers backed by an SDK that supports the `remote_control`
+   * control request implement this.
+   */
+  enableRemoteControl?(enabled: boolean, name?: string): Promise<unknown>;
   cleanup(): void;
   rewind(userMessageId: string, assistantMessageId: string, mode?: ChatRewindMode): Promise<ChatRewindResult>;
   setApprovalCallback(callback: ApprovalCallback | null): void;
--- a/src/core/runtime/types.ts
+++ b/src/core/runtime/types.ts
@@ -3,6 +3,7 @@ import type { CanvasSelectionContext } from '../../utils/canvas';
 import type { EditorSelectionContext } from '../../utils/editor';
 import type {
   ApprovalDecision,
+  ChatMessage,
   Conversation,
   ExitPlanModeCallback,
   ImageAttachment,
@@ -108,6 +109,14 @@ export interface ChatTurnMetadata {
 export interface AutoTurnResult {
   chunks: StreamChunk[];
   metadata: ChatTurnMetadata;
+  /**
+   * User message that triggered this auto-turn, when it originated outside the
+   * local UI (e.g. a remote-control turn sent from claude.ai). The SDK stream
+   * does not echo such prompts, so the runtime sources it from the transcript
+   * for the renderer to show before the assistant reply. Undefined for local
+   * background turns (e.g. subagent notifications).
+   */
+  precedingUserMessage?: ChatMessage;
 }
 
 export type AutoTurnCallback = (result: AutoTurnResult) => void | Promise<void>;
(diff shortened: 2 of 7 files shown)


YishenTu:
Review notes after looking at the Remote Control flow:

## Recommendation

Given the preferred direction is **reload-only**, I recommend making that mode explicit and consistent: do not live-render remote-control turns into Claudian chat state/DOM. Keep live auto-turn rendering only for existing background task/subagent notification flows, and recover remote-control turns from Claude JSONL history when the user reloads/switches the chat.

One requirement still remains: Remote Control should only be enabled for a tab that already has durable Claudian conversation/session metadata, or enabling Remote Control should create/bind that metadata up front. Otherwise a remote-first turn on a blank tab can have no Claudian conversation id/session id for history reload to find later.

## Findings

1. **P1: Current patch is a broken middle state** (`src/features/chat/tabs/Tab.ts:1803`)

   The patch live-renders remote-control turns by appending the reconstructed user prompt and assistant response directly into `tab.state`/DOM, but it does not persist through `ConversationController.save(...)`. That leaves conversation metadata stale (`lastResponseAt`, preview/title state, provider state), and in some same-session flows can restore stale `conversation.messages` after switching/reloading the chat.

   If the desired model is reload-only, the fix should be to classify/suppress remote-control auto-turn chunks from the live `AutoTurnCallback` path and rely on provider history hydration instead. If live mirror is kept instead, this needs to go through the normal conversation save boundary after rendering.

2. **P2: Production console logging** (`src/main.ts:195`)

   The command logs the raw `remote_control` response with `console.log`. Project instructions prohibit `console.*` in production code, and this response likely includes pairing/session URL data. Please remove it or route it through an intentional debug surface.

## Verification

- `npm run typecheck` passed.
- Focused tests passed: `npm run test -- tests/unit/providers/claude/runtime/ClaudianService.test.ts tests/unit/features/chat/tabs/Tab.test.ts --runInBand`
- Touched-file lint passed.

liqing-ustc:
Thanks for the detailed review — both findings were valid.

**P2:** Removed the `console.log` in `src/main.ts`; the `Notice` already surfaces the result to the user.

**P1:** Kept the live mirror but routed it through the save boundary. `renderAutoTriggeredTurn` now calls `conversationController.save(true)` after rendering a remote/auto turn, so conversation metadata (`lastResponseAt`, title/preview, provider state) stays consistent and a later switch/reload won't restore stale messages. This also covers the binding requirement: `save()` lazily creates and binds the conversation with the SDK `sessionId` when the tab has none yet (`ConversationController.ts:398-405`), so a remote-first turn on a blank tab is recoverable from history on reload.

Verified: `npm run typecheck`, full `npm run test` (240 suites / 5677 tests), and lint pass.

PR #746, opened 2026-06-06
feat(pi): add search filter to discovered models list

madhusudan-kulkarni:
## What

Adds a filter input to the Pi discovered models list in settings.
Type to narrow by model name, encoded ID, or alias. Displays a
count when active (e.g., "3 of 12 models shown").

## Why

With many discovered models, finding and toggling a specific one
requires scrolling through the full list. A filter makes this
instant.

## Changes

- `src/providers/pi/ui/PiSettingsTab.ts`: add filter state, input,
  and filtered rendering

## Validation

- `npm run typecheck` — passes
- `npm run lint` — passes
- `npm run test` — 5677/5677 pass
- `npm run build` — succeeds
- Manual: Pi settings → Models → type in filter → list narrows
  correctly, toggle and alias still work

### DIFF
--- a/src/providers/pi/ui/PiSettingsTab.ts
+++ b/src/providers/pi/ui/PiSettingsTab.ts
@@ -97,14 +97,40 @@ export const piSettingsTabRenderer: ProviderSettingsTabRenderer = {
 
     new Setting(container).setName('Models').setHeading();
 
+    let searchQuery = '';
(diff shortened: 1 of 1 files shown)


YishenTu:
i made the pi model list unified with opencode's style in #760 , see if you like that? i think a consistent style is better than the current scattered one.

madhusudan-kulkarni:
Looks great, the unified style is cleaner. Thanks for picking this up.

PR #747, opened 2026-06-06
feat: support drag-and-drop files and folders in chat view input

karmonlong:
Supports dragging files and folders from Obsidian's internal File Explorer (as well as external files from the OS) and dropping them directly into the Claudian chat view input area to automatically add them as @ mentions. Normalizes relative paths and handles folders (appends trailing slash). Includes full unit test coverage.


xueyang-dev:
I opened #830 to describe the user workflow from an Obsidian user perspective. #747 looks very close to the desired drag-and-drop behavior. I would be happy to help test or refine the UX if this direction is still welcome.

PR #753, opened 2026-06-07
feat: right-sidebar session switcher

mfassaie:
## Summary

Adds a companion **right-sidebar session switcher** — an Obsidian view that lists open tabs and recent conversations and drives the active chat view's `TabManager`. It's offered as a new **tab bar position** so switching can live in the sidebar instead of the in-chat badges.

This is primarily intended for the use case where the Claudian chat panel is opened in the **main editor window** (rather than in a sidebar). With chat in the main window, the right sidebar hosts the switcher and follows the active chat tab, while still freeing up for the document Outline on non-chat tabs.

No changes to runtimes, providers, or the conversation model — this is a UI/coordination layer over the existing `TabManager` and conversation list.

## What's included

- **New view + entry points**: `claudian-sessions-view` (`SessionsView`) hosting a plain-DOM `SessionsPanel`; a ribbon icon and an `open-sessions-panel` command.
- **SessionsPanel**: two sections — *Open* (from `getTabBarItems()`, with active/streaming/attention state and a provider dot; row → `switchToTab`, close → `closeTab(id, force)`) and *Recent* (`getConversationList()` minus already-open conversations; row → `openConversation(id, { preferNewTab, activate })`), plus a *New session* button. Resolves the target chat view via a new `getActiveChatView()`, opening the chat first if none exists.
- **Sync seam**: a lightweight plugin emitter (`onSessionsChanged` / `notifySessionsChanged`) fired from the existing `TabManager` callbacks and conversation CRUD, so the panel stays current.
- **New tab bar position `right-sidebar`**: hides the in-chat tab badges, auto-opens the panel, and makes the sidebar follow the active chat tab (via `active-leaf-change`) while leaving the sidebar free for the document Outline on non-chat tabs.

## Tests

- `SessionsPanel` renderer: open-tabs section with active highlight, recent section that excludes already-open conversations, and the right calls on click/close/new.
- Plugin emitter: subscribe receives notifications; unsubscribe stops delivery; a failing listener is isolated.

`npm run typecheck`, `npm run lint`, and the new unit suites pass.

### DIFF
--- a/src/core/types/chat.ts
+++ b/src/core/types/chat.ts
@@ -11,6 +11,9 @@ export interface ForkSource {
 /** View type identifier for Obsidian. */
 export const VIEW_TYPE_CLAUDIAN = 'claudian-view';
 
+/** View type identifier for the companion right-sidebar session switcher. */
+export const VIEW_TYPE_CLAUDIAN_SESSIONS = 'claudian-sessions-view';
+
 /** Supported image media types for attachments. */
 export type ImageMediaType = 'image/jpeg' | 'image/png' | 'image/gif' | 'image/webp';
 
--- a/src/core/types/index.ts
+++ b/src/core/types/index.ts
@@ -11,6 +11,7 @@ export {
   type StreamChunk,
   type UsageInfo,
   VIEW_TYPE_CLAUDIAN,
+  VIEW_TYPE_CLAUDIAN_SESSIONS,
 } from './chat';
 export { type ProviderId } from './provider';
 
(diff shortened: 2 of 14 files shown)


mfassaie:
@YishenTu the failing `test` check is fixed in d473535.

The integration suite's `onload` tests threw `TypeError: this.app.workspace.on is not a function` because this PR registers a `workspace.on('active-leaf-change')` listener in `onload()`, and the test mocks didn't expose `workspace.on` / `Plugin.registerEvent`. I've added those to the mocks (`tests/integration/main.test.ts` and `tests/__mocks__/obsidian.ts`).

Verified locally: integration suite passes 205/205, lint and typecheck clean. The latest run is in `action_required` state — would you mind approving the workflow run so CI can confirm green? Thanks!

PR #758, opened 2026-06-07
fix: improve fast mode icon contrast

zacandmolly:
## Summary
- Improves the active Codex fast-mode icon contrast in the toolbar.
- Keeps the existing Lucide zap icon, but renders the active state with a white fill and black outline so it stays visible on dark and light Obsidian themes.
- Adds a subtle active button background/border to make the fast-mode state easier to identify.

## Test Plan
- npm run test -- --selectProjects unit --runTestsByPath tests/unit/features/chat/ui/InputToolbar.test.ts
- npm run build
- npm run typecheck
- npm run lint

### DIFF
--- a/src/style/toolbar/service-tier-toggle.css
+++ b/src/style/toolbar/service-tier-toggle.css
@@ -15,7 +15,7 @@
   color: var(--text-muted);
   background: transparent;
   cursor: pointer;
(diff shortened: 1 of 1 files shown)


YishenTu:
i like the way it is.

PR #774, opened 2026-06-11
fix: fix the compatibility error for working with codex on WSL

Tonanguyxiro:
Fix the error for working with codex on WSL2:

- Add selections for WSL1 and WSL2
- Automatically get WSL distribution for selection
- Use login terminal for starting codex in WSL
- More detailed error output

Now it works well on WSL2


YishenTu:
Blocking issue in `manifest.json`: this PR changes the Obsidian plugin id from `realclaudian` to `claudian-dev` and changes only the manifest version to `2.0.22.1` while `package.json` remains `2.0.22`.

Changing the plugin id means existing installs will not update as the same plugin; Obsidian will treat it as a different plugin/settings namespace. Please revert the id change and keep the manifest version in sync through the normal release/version path, unless this PR is intentionally publishing a separate dev plugin.

Tonanguyxiro:
Thanks for remind me, I modify the ID for test, I will resolve such conflicts.

PR #782, opened 2026-06-12
feat: add Claude WSL support and unify WSL utilities for Codex

Tonanguyxiro:
Add WSL 1 and WSL 2 support for the Claude provider, while consolidating shared WSL distribution discovery and path mapping used by both Claude and Codex.

## Changes

- Add configurable WSL 1/WSL 2 execution for Claude on Windows.
- Support Windows drive paths and WSL UNC workspaces.
- Load Claude settings, history, plugins, and agents from WSL.
- Map Claude tool paths between Linux and Windows.
- Extract shared WSL distro discovery and path mapping utilities.
- Update Codex execution target and path mapping to use the shared utilities.
- Preserve existing Codex WSL APIs through compatibility exports.
- Add WSL execution, spawning, settings, and path mapping tests.

image

image


YishenTu:
Review findings:

1. **P1: Invalid/incomplete Claude WSL settings can break conversation switching.**  
   `src/providers/claude/history/ClaudeConversationHistoryService.ts:389` resolves WSL execution paths before checking whether the conversation has SDK sessions, and `src/main.ts:632` does not catch hydration errors. A temporary state like “WSL 2 selected, distro not selected yet” can throw from `workspace.resolveExecutionContext()` and prevent opening Claude conversations. History hydration/deletion should fail soft, or defer WSL path resolution until there is a session to load.

2. **P2: Claude WSL Linux `.cjs` CLI paths can be rejected by host-side Node validation.**  
   The main query path and cold-start path call `getMissingNodeError()` before resolving whether execution is WSL (`src/providers/claude/runtime/ClaudeChatRuntime.ts:1136`, `src/providers/claude/runtime/claudeColdStartQuery.ts:67`). But the WSL spawn path intentionally runs Node-backed CLI wrappers with WSL’s `node` (`src/providers/claude/runtime/customSpawn.ts:38`). A Windows host without `node.exe` can reject `/home/user/.../cli-wrapper.cjs` even though WSL has Node and would run it correctly. Resolve execution context first and skip host Node validation for WSL, or validate inside WSL.

3. **P2: Streamed Claude tool-use paths are not mapped back to host paths.**  
   Full assistant messages map tool input paths through `toHostPath` (`src/providers/claude/stream/transformClaudeMessage.ts:419`), but streamed `content_block_start` and `input_json_delta` tool-use events emit the raw target input (`src/providers/claude/stream/transformClaudeMessage.ts:523`). In WSL mode, running tool cards/file links can show `/mnt/c/...` or `/home/...` instead of Windows/UNC host paths until, or unless, a later full message replaces them.

Verification run on PR head `9a807b3`:

- `npm run typecheck` passed.
- Targeted WSL/runtime/settings tests passed.
- Claude runtime/storage suites passed after linking the existing dependency install into the throwaway worktree.
- One attempted command referenced `tests/unit/providers/claude/history/ClaudeConversationHistoryService.test.ts`, which does not exist in this PR.

PR #807, opened 2026-06-23
fix: focus chat input after opening Claudian

### LINKED ISSUE #739: Input not focused when Claudian opens in Main editor tab via command/hotkey (persists in 2.0.21, cf. #362)
## Summary

When "Open Claudian in" is set to **Main editor tab**, opening Claudian via the command/hotkey (`realclaudian:open-view`) opens the chat panel but **does not focus the input field**. You have to click the input once before you can type. This is the same symptom as the (closed) #362, and it still reproduces in the latest release.

## Environment

- Claudian: **2.0.21** (latest)
- Obsidian: Desktop, macOS (Darwin)
- Setting: Display → **Open Claudian in = Main editor tab**

## Steps to reproduce

1. Settings → Claudian → Display → set **Open Claudian in** to **Main editor tab**.
2. Close any existing Claudian panel so a fresh one is created.
3. Trigger **Claudian: Open view** (e.g. via a hotkey like ⌘U, or the command palette).
4. Claudian opens in a main editor tab.
5. Start typing.

## Expected

The chat input is focused on open, so you can type immediately.

## Actual

Focus stays on the editor/tab; keystrokes don't go to the chat input. You must click the input field first.

## Notes

- Opening Claudian in the **sidebar** focuses the input fine — only the **Main editor tab** path misses focus.
- This matches #362 (closed), so it looks like a regression / still-open limitation for the main-tab case. Auto-focusing the input on creation in a main editor tab would fix it.
- Possibly separate, but noticed alongside: the **`Claudian: New session`** command appears to be a no-op in 2.0.21 (nothing happens), while `Open view` and `Close current tab` work. Happy to file that separately if useful.

snvtac:
## Summary
- focus the active Claudian chat input after `Open chat view` reveals an existing or newly-created leaf
- add `ClaudianView.focusActiveInput()` so the plugin entrypoint uses a view-level API instead of reaching into tab DOM internals
- retry focus briefly while a cold-open tab finishes restoring, and skip delayed focus if another leaf becomes most recent

## Root cause
`activateView()` revealed the Claudian leaf but never moved keyboard focus into the chat textarea, so commands/hotkeys could open the view while subsequent typing still went to the previously focused editor.

## Validation
- `npm run typecheck`
- `npm run lint`
- `npm run test` (242 suites, 5760 tests)
- `npm run build`

Fixes #739

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -41,10 +41,14 @@ import { buildCursorContext } from './utils/editor';
 import { revealWorkspaceLeaf } from './utils/obsidianCompat';
 import { getVaultPath } from './utils/path';
 
+const INPUT_FOCUS_RETRY_DELAY_MS = 100;
+const INPUT_FOCUS_MAX_ATTEMPTS = 3;
+
 function isClaudianView(value: unknown): value is ClaudianView {
   return !!value
     && typeof value === 'object'
-    && typeof (value as { getTabManager?: unknown }).getTabManager === 'function';
+    && typeof (value as { getTabManager?: unknown }).getTabManager === 'function'
+    && typeof (value as { focusActiveInput?: unknown }).focusActiveInput === 'function';
 }
 
 export default class ClaudianPlugin extends Plugin {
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -210,9 +214,32 @@ export default class ClaudianPlugin extends Plugin {
 
     if (leaf) {
       await revealWorkspaceLeaf(workspace, leaf);
+      this.focusInputAfterReveal(leaf);
     }
   }
 
+  private focusInputAfterReveal(leaf: WorkspaceLeaf, attempt = 1): void {
+    window.setTimeout(() => {
+      const mostRecentLeaf = this.app.workspace.getMostRecentLeaf?.();
```
YishenTu, line 223: This guard makes the focus fix ineffective for the default sidebar placement. `getMostRecentLeaf()` without a root searches the root split/pop-outs, not sidebar leaves, so after `revealLeaf()` opens Claudian in the default right sidebar it can still return the previously active editor leaf. That makes `mostRecentLeaf !== leaf` true and returns before `focusActiveInput()` runs. The cancellation check should not use this root-leaf API for sidebar targets, or should only apply it when the Claudian leaf is in the root area. Please add coverage where Claudian opens in `right-sidebar` and `getMostRecentLeaf()` returns the previous editor leaf; the input should still be focused.

YishenTu (commented):
I found one behavioral issue with the focus guard. The PR’s touched tests pass locally, but they do not cover Obsidian’s sidebar behavior for `getMostRecentLeaf()`.

PR #813, opened 2026-06-24
feat: add in-app prompt library

thy950523:
## Summary

- Local prompt library: adds a 📖 button to the input toolbar that opens a popover for managing preset prompts — create, edit, and delete, with live search by name and content.
- Quick insert: selecting a prompt fills the chat input (still editable before sending); clicking outside the popover dismisses it.
- Provider-neutral storage: prompts persist to .claudian/prompts.json via PromptLibraryStorage (with structural validation on load/save), shared across Claude and Codex tabs. All 10 locales are covered.

image

## Test Plan

- [x] npm run typecheck && npm run lint && npm run test && npm run build — all green (5757 tests passing)
- [x] New unit tests: PromptLibraryStorage (8), filterPrompts (3), PromptLibraryPanel (1)
- [x] Manual verification: toolbar 📖 button → popover opens → create/edit/delete → click to insert into input → search filtering → click-outside-to-close → persistence across Obsidian restarts via .claudian/prompts.json → works on both Claude and Codex tabs

### DIFF
--- a/docs/plans/2026-06-24-prompt-library.md
+++ b/docs/plans/2026-06-24-prompt-library.md
@@ -1,0 +1,909 @@
+# Prompt Library Implementation Plan
+
+> **For Claude:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task.
+
+**Goal:** Add an in-app local "prompt library" — create, edit, delete preset prompts from a toolbar panel, and insert a chosen prompt into the chat input box (editable before sending).
+
+**Architecture:** A provider-neutral `PromptLibraryStorage` reads/writes `.claudian/prompts.json` via the existing `VaultFileAdapter`, exposed through `SharedAppStorage`. A toolbar button (added to `createInputToolbar`) opens a popover `PromptLibraryPanel` that lists/searches prompts and supports inline CRUD. Selecting a prompt calls an `onInsert(content)` callback that fills the tab's `textarea`.
+
+**Tech Stack:** TypeScript, Obsidian plugin APIs (DOM `createDiv`/`setIcon`, `Notice`, `Modal`), Jest unit tests, modular CSS under `src/style/`.
+
+**Confirmed design decisions (from brainstorm):**
+- Selecting a prompt **replaces** the current input draft (then editable before send).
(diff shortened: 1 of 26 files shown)


YishenTu:
how is it different from skills/commands?

thy950523:
**Short answer: skills are automated routines the model runs; the prompt library is a lightweight notebook of plain-text snippets that just save me retyping. They overlap on "reusable text," but not on intent.**

A bit of context on why I built it. I'm a fairly new Obsidian user, and my main use case is reading — I do a lot of web clippings, and what I want most is to quickly pull the core argument out of an article I've saved. Claudian living in my sidebar has become where I do that. I keep a few personal prompts for digesting those clippings, and the friction was simple: every time I wanted one I had to go dig it out of wherever I'd stashed it and paste it in. The library started as just wanting those one click away.

On why I didn't make them skills — and I'd value your take, since you know this codebase far better than I do. My mental model of a skill is something **heavier**: an automated workflow, like "summarize → turn into a knowledge card → file into a specific folder," a repeatable pipeline the model runs, often with tool limits or a fixed output shape. What I need is much lighter — plain-text tidying and summarization, **me staying in the loop**, editing the filled-in text before I send. It felt wrong to wrap that in skill machinery when there's no real flow to automate.

The second reason is **portability across providers**. As I understand it, a skill is bound to its provider — Claude skills live in `.claude/skills`, Codex skills in `.codex/skills`, and a skill I create on a Claude tab simply isn't visible when I switch to Codex. The prompt library is one `.claudian/prompts.json` shared by every tab, so the same snippet follows me regardless of which provider I'm talking to. Skills can't do that.

So to sum up the difference the way I now see it: a **skill** is a structured, provider-bound routine for automating a repeatable task; a **prompt** is a provider-neutral snippet for quickly seeding a message I'm about to write myself. Two different jobs, and I think there's room for both.

Thanks for pushing on this — it genuinely made me think about the boundary between the two a lot more clearly.

YishenTu:
you are over-thinking about skills/prompts/commands, they can be as simple as reusable instruction despite your mental model.

you can use $ARGUMENTS inside a skills to achieve versatility, and symlink to share across providers.

PR #840, opened 2026-06-29
feat(i18n): add Polish translation

Leon135:
Added complete Polish translation

Im native speaker


Leon135:
Should be good to go

YishenTu:
i have updated the whole i18n setup, this pr needs update, it will be merged once resolve.

Issue #846, opened 2026-07-01
Sonnet 5 support


astoeffer:
Hi,

is the use of Sonnet 5 possible right now?
And will Fable 5 be supported after redeploying through Anthropic?

Thanks
Andreas

lpburke86:
Sonnet 5 access would be amazing sooner than later. I have a project I’m on right now that 4.6 won’t handle, so I’m stuck using Opus for it at twice the cost.

RomeDevv:
Sonnet 5 is already usable today — no picker change was needed for that part. Claudian's generic aliases (`haiku`/`sonnet`/`opus`) get resolved server-side by the Claude Code CLI to the latest model in that tier, so selecting "Sonnet" already targets Sonnet 5.

What's actually missing: the `xhigh` effort/thinking level was only allow-listed for Opus 4.7+, so it stayed hidden when Sonnet was selected even though Sonnet 5 supports it. That's fixed by #853 (pending review) — once merged, selecting Sonnet will show `XHigh` as an available effort option.

(#853 also adds Fable 5 as a picker entry, which covers the other half of this issue.)

PR #847, opened 2026-07-01
fix: allow xhigh effort level for Sonnet 5

RomeDevv:
## Summary
- Claudian's model picker uses generic aliases (`haiku`/`sonnet`/`opus`) that the Claude Code CLI resolves server-side, so the "Sonnet" option already targets Sonnet 5 with no picker changes needed.
- `supportsXHighEffort()` only allow-listed Opus 4.7+, so the `xhigh` effort level stayed hidden for Sonnet even though Sonnet 5 is the first Sonnet-tier model to support it.
- Extends `supportsXHighEffort()` to recognize the built-in `sonnet`/`sonnet[1m]` aliases and explicit `claude-sonnet-5`+ ids, while still excluding `claude-sonnet-4-x`, which doesn't support `xhigh`.
- Updates two stale comments and the related unit tests accordingly.

Addresses the Sonnet 5 support question raised in #846.

Co authored with Sonnet 5 :P

## Test plan
- [x] `npm run test -- --selectProjects unit` — all previously-passing tests still pass; no new failures introduced (verified against a clean `main` baseline, which has 13 pre-existing unrelated failing suites on this Windows environment)
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run build`

### DIFF
--- a/src/providers/claude/runtime/ClaudeQueryOptionsBuilder.ts
+++ b/src/providers/claude/runtime/ClaudeQueryOptionsBuilder.ts
@@ -301,8 +301,8 @@ export class QueryOptionsBuilder {
   ): void {
     const effortLevel = resolveEffortLevel(model, settings.effortLevel);
     options.thinking = { type: 'adaptive' };
-    // SDK runtime accepts `xhigh` on Opus 4.7+ and silently falls back to
-    // `high` elsewhere, but its type definition lags our local EffortLevel.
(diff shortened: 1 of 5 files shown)


RomeDevv:
Superseded by #853, which combines this with #850 into one branch to avoid a merge conflict in supportsXHighEffort() for whoever reviews it. Closing in favor of the combined PR.

PR #855, opened 2026-07-03
fix: preserve image attachment data in conversation history

SkyNotSilent:
## Summary

- Keep image attachment base64 data on conversation messages after metadata saves
- Add an integration regression test for `updateConversation` preserving pasted image data
- Align the main update flow with `ImageAttachment.data` being the single source of truth for chat UI image rendering

## Why

Pasted and dropped images render from `ChatMessage.images[].data` in the history UI and full-image modal. `updateConversation` was clearing that data after saving metadata, while metadata does not include messages, so opening the conversation later could leave the image source as an empty data URI.

This intentionally preserves image payloads on the in-memory conversation messages because the current chat UI contract treats `ImageAttachment.data` as the display source of truth. If memory pressure becomes a concern later, the safer follow-up would be to introduce a separate durable attachment store and update the renderer to resolve from that store, rather than blanking the only UI-readable payload after metadata saves.

Related prior work searched:
- #62 added image rendering in chat UI
- #106 moved session metadata toward SDK-native storage
- #226 added fork handling, including the previous special case around image data

I did not find an existing issue or PR that directly covers pasted images failing to open from history.

## Test plan

- `npm run test -- tests/integration/main.test.ts --runInBand`
- `npm run typecheck`
- `npm run lint`
- `npm run test -- --runInBand`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -684,18 +684,6 @@ export default class ClaudianPlugin extends Plugin {
     await this.storage.sessions.saveMetadata(
       this.storage.sessions.toSessionMetadata(conversation)
     );
-
-    // Clear image data from memory after save (data is persisted by SDK).
-    // Skip for pending forks: their deep-cloned images aren't in SDK storage yet.
-    if (!ProviderRegistry.getConversationHistoryService(conversation.providerId).isPendingForkConversation(conversation)) {
-      for (const msg of conversation.messages) {
-        if (msg.images) {
-          for (const img of msg.images) {
-            img.data = '';
-          }
-        }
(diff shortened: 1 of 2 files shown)


YishenTu:
Closing this in favor of #865. I have a broader fix there that covers image rehydration across all providers we checked: Claude, Codex, Pi, and OpenCode. It also keeps the original PR 855 behavior fix by preserving image base64 data, and adds the Codex user-message cleanup so generated <image ...></image> transport tags are not rendered.

PR #882, opened 2026-07-08
feat: account usage limits meter (5-hour session + weekly windows)

mrahar:
## Summary

Adds a **usage limits meter** to the chat input toolbar, mirroring the Claude app's `/usage` display. Pro/Max subscribers can now see how much of their 5-hour session and weekly quotas they've consumed without leaving Obsidian.

## What it does

- A small **gauge icon** appears in the input toolbar next to the context meter.
  - After usage has been fetched, the icon turns orange at ≥70% session usage and red at ≥90%, and its tooltip shows the current session percentage.
- Clicking it opens a **popover** with progress bars for:
  - **Current session (5h)** — with reset time
  - **Current week (all models)** — with reset time
  - **Current week (top models)** (`weekly_scoped`, shown when the API reports it)
- **No background network activity.** Usage is fetched **only when the user opens the popover** — the plugin never contacts the network on its own. Responses are cached for 60s so repeated opens don't re-fetch.
- The popover is **dynamically positioned to stay inside the view container**, so it doesn't get clipped at the panel edge in the narrow side panel (measured against `.view-content`, width-capped to the panel, shifted horizontally to fit).
- A new **setting** (Claude tab → Models → "Usage limit reset display") lets users choose the reset time format:
  - *Time remaining* (default): `Resets in 4h 20m`
  - *Exact time*: `Resets at 3:49 AM`

## How it works

- **`src/core/usage/UsageLimitsService.ts`** — reads the Claude Code OAuth access token from the OS credential store (macOS Keychain `Claude Code-credentials`, with `~/.claude/.credentials.json` as fallback for Linux/Windows) and calls the official `https://api.anthropic.com/api/oauth/usage` endpoint with the `oauth-2025-04-20` beta header. Responses are cached for 60s with in-flight request deduplication. The token is never logged, displayed, or sent anywhere except api.anthropic.com. Uses Obsidian's `requestUrl` (no CORS issues).
- **`src/features/chat/ui/UsageLimitsMeter.ts`** — toolbar component + popover. Fetches on demand (popover open) only — no timers, no polling. Reads the display preference from the existing toolbar settings snapshot (no new plumbing). Clamps the popover within the plugin's view bounds.
- **`src/providers/claude/settings.ts`** — new `usageLimitsResetDisplay: 'remaining' | 'absolute'` provider setting (default `'remaining'`), normalized like existing fields.
- **`src/style/toolbar/usage-limits.css`** — theme-safe styling using Obsidian CSS variables only (works in both light and dark themes).
- **i18n** — strings added to all 10 locales.

## Privacy / network behavior

- The plugin makes **zero unsolicited network requests**. The only call to `/api/oauth/usage` happens in direct response to the user clicking the gauge icon.
- That endpoint reports account usage; it does **not** consume the inference rate limit.

## Error handling

- No credentials found → friendly hint to sign in with `claude` in a terminal.
- Endpoint failure (e.g. API-key-only accounts without a subscription) → graceful message inside the popover.

## Testing

- `tsc --noEmit` clean, `eslint` clean (0 warnings — popover uses `ownerDocument` for popout-window compatibility, and avoids static inline-style assignments per `obsidianmd/no-static-styles-assignment`).
- All existing tests pass: 243 suites / 5797 tests.
- Manually tested on macOS (Obsidian desktop, dark + light themes, narrow side panel) against a live Pro account: percentages and reset times match the official `/usage` command output exactly, and the popover stays fully inside the panel.

## Notes

- Desktop-only, same as the plugin itself (uses `child_process`/`fs` for credential lookup, consistent with existing code).
- The displayed values come verbatim from Anthropic's endpoint — no client-side estimation.

---
This feature was built through **vibe coding** — an AI pair-programming session with Claude, with human direction, review, and manual testing throughout.


YishenTu:
i recommend [Codexbar](https://codexbar.app/).

PR #897, opened 2026-07-10
Fix Windows WSL distro discovery and startup failure

### LINKED ISSUE #895: Plugin fails to load on Windows: "Unable to determine the WSL distro" despite a default distro being configured (v2.0.31)
## Environment

- Claudian v2.0.31 (installed/updated via BRAT)
- Obsidian on Windows 11 Pro (10.0.26200)
- WSL2 installed, `Ubuntu` is set as the **default** distro (`wsl --status` confirms "Default Distribution: Ubuntu")
- Codex provider configured with `installationMethod: "wsl"`, no explicit distro override set

## What happens

After BRAT auto-updated to v2.0.31, the plugin fails to enable entirely. Toggling it in Community Plugins immediately reverts, and the console shows:

```
Plugin failure: realclaudian Error: Unable to determine the WSL distro. Set WSL distro override or configure a default WSL distro.
    at buildCodexLaunchSpec (plugin:realclaudian:71804:11)
    at resolveCodexAppServerLaunchSpec (plugin:realclaudian:71869:10)
    at CodexModelDiscoveryService.discoverModels (plugin:realclaudian:72739:24)
    at Object.refreshModelCatalog (plugin:realclaudian:75079:43)
    at createCodexWorkspaceServices (plugin:realclaudian:75109:35)
    at async ProviderWorkspaceRegistry.initializeAll (plugin:realclaudian:25210:35)
    at async ClaudianPlugin.onload (plugin:realclaudian:107672:5)
```

The setup worked fine before the update (previous version, same machine, same WSL config).

## Two issues, as I see it

1. **Default distro detection fails** even though a default WSL distro is configured. Possibly related to parsing `wsl.exe` output, which is UTF-16LE encoded on Windows.
2. **The error is thrown during `onload` and takes the whole plugin down.** Since the recommended fix ("Set WSL distro override") is a plugin setting, and the settings UI is unreachable when the plugin won't load, users are locked out. A failed Codex model discovery should probably degrade gracefully (disable the provider, show a notice) instead of aborting plugin load.

## Workaround

Manually editing `.claudian/claudian-settings.json` and setting the distro explicitly under the Codex provider config fixes it:

```json
"wslDistroOverridesByHost": {
  "device:<device-id>": "Ubuntu"
}
```

After that, the plugin loads normally again.

## Related

Possibly related to #471 / PR #860 — same code area (Codex installation method / WSL launch logic), but the opposite platform: this happens on real Windows with a real WSL setup, where the default-distro detection fails and, unlike #471, the thrown error aborts the entire plugin `onload`.

YishenTu:
## Summary

- Decode raw `wsl.exe --list --verbose` output as UTF-16LE or UTF-8 so the configured default distro is detected reliably.
- Contain Codex launch-spec failures as model-discovery diagnostics so provider initialization cannot abort plugin startup.
- Add regression coverage for both Windows encoding variants and pre-process discovery failures.

Fixes #895.

### DIFF
--- a/src/providers/codex/runtime/CodexExecutionTargetResolver.ts
+++ b/src/providers/codex/runtime/CodexExecutionTargetResolver.ts
@@ -40,8 +40,43 @@ export function inferWslDistroFromWindowsPath(hostPath: string | null | undefine
   return match?.[1] || undefined;
 }
 
-export function parseDefaultWslDistroListOutput(output: string): string | undefined {
-  for (const line of output.replace(/\uFEFF/g, '').split(/\r?\n/)) {
+function looksLikeUtf16Le(output: Buffer): boolean {
+  const sampleLength = Math.min(output.length - (output.length % 2), 512);
+  if (sampleLength < 4) {
+    return false;
+  }
+
+  let evenNullBytes = 0;
+  let oddNullBytes = 0;
+  for (let index = 0; index < sampleLength; index += 2) {
+    if (output[index] === 0) {
+      evenNullBytes += 1;
+    }
+    if (output[index + 1] === 0) {
+      oddNullBytes += 1;
+    }
+  }
+
+  const bytePairs = sampleLength / 2;
+  return oddNullBytes / bytePairs >= 0.2 && oddNullBytes > evenNullBytes * 2;
+}
+
+function decodeWslListOutput(output: string | Buffer): string {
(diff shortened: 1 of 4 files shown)


gomerpyle:
Thank you for the quick fix and for addressing both issues! I’ll test it once the next release is available.

Issue #513, opened 2026-04-16
Stop:Callback blocks session exit after foreground Agent completions


benjibradley:
## Environment
- Claudian v2.0.2
- Obsidian on macOS (Apple Silicon)
- Claude Code CLI spawned via stdio JSON bridge

## Description

When foreground agents are spawned via the `Agent` tool during a Claudian session, an internal `Stop:Callback` hook blocks session termination indefinitely — even after all agents have completed and returned results.

```
Stop:Callback hook blocking error from command: "callback": Background subagents are still running.
Use `TaskOutput task_id="..." block=true` to wait for their results before ending your turn.
```

The error fires on every stop attempt. Only escape is `/clear` or closing the session.

## Repro

1. Open Claudian in Obsidian sidebar
2. Run a task that spawns 2+ **foreground** agents via the `Agent` tool (not `run_in_background`)
3. Wait for all agents to return results inline
4. Try to end the session (or let a custom Stop hook fire)

## Investigation

Thoroughly checked — the issue is NOT from user-configured hooks:

- Custom Stop hook (`session-capture.py`) fires once then allows — working correctly
- `TaskOutput` returns "No task found" for all agent IDs
- No agent processes, lock files, or tracking state exists anywhere
- `~/.claude/tasks/` empty, `/tmp` has no agent tracking files
- GrowthBook feature flags checked — no subagent flags control Stop behavior

The `Stop:Callback` is an internal mechanism (`type: "callback"` vs user hooks' `type: "command"`). It's registered by the runtime when agents are spawned but never deregistered when foreground agents complete.

## Hypothesis

This may be a Claude Code CLI runtime issue (filed as [anthropics/claude-code#49071](https://github.com/anthropics/claude-code/issues/49071)), but the Claudian stdio JSON bridge is a possible contributing factor. When the Agent tool spawns subagents, both the CLI and the Claudian bridge layer may track agent lifecycle. If Claudian's tracking doesn't process the completion signal, a stale callback persists and blocks Stop indefinitely.

Filing here because the bridge between Claudian and the CLI is where the double-tracking could occur.

## Related

- [anthropics/claude-code#49071](https://github.com/anthropics/claude-code/issues/49071) — same bug filed against the CLI runtime

YishenTu:
try latest version 2.0.3

Issue #493, opened 2026-04-14
Stop hook infinite loop: falsely reports background subagents still running after foreground Agent completes


11-Liuc:
## Bug Description

After a **foreground** Agent tool call completes successfully, Claudian's built-in Stop callback hook keeps returning `"Background subagents are still running"`, causing Claude to enter an **infinite loop** — unable to end its turn, burning tokens on every cycle.

## Steps to Reproduce

1. Start a conversation in Claudian
2. Claude uses `Agent` tool with `run_in_background: false` (foreground)
3. Agent returns results normally
4. Claude tries to end its turn → **Stop hook blocks** → infinite loop

**This does not happen in terminal Claude Code** — only through Claudian.

## Environment

- Claudian: v2.0.2
- Obsidian: v1.8.9
- OS: macOS (Apple Silicon)
- Theme: Minimal

## Root Cause Analysis

Traced through `main.js`:

1. **`createStopSubagentHook`** (L57896) registers a Stop callback that checks `getState().hasRunning`

2. **`hasRunningSubagents()`** (L76256) returns `true` when `pendingAsyncSubagents.size > 0 || activeAsyncSubagents.size > 0`

3. **Suspected issue 1 — catch branch defaults to `true`** (L57903):

```javascript
try {
    hasRunning = getState().hasRunning;
} catch (e) {
    hasRunning = true;  // ← defaults to blocking on any error
}
```

If `getState()` throws (e.g., tab/service lifecycle issue), the hook permanently blocks.

4. **Suspected issue 2 — foreground agent state leaking into async maps**: A foreground agent may be temporarily added to `pendingAsyncSubagents` (L76340) but the completion event doesn't trigger `delete` from the map, leaving it non-empty.

## Suggested Fixes

1. Change the catch branch to `hasRunning = false` — better to miss a check than to cause an infinite loop
2. Add debug logging in the Stop hook to output current `pendingAsyncSubagents` and `activeAsyncSubagents` contents
3. Verify that foreground Agent lifecycle never writes to the async state maps, or ensure proper cleanup if it does

## Impact

Each blocked cycle consumes tokens and costs money. In one conversation, this triggered 20+ consecutive forced loops before the user manually interrupted.

## Related Issues in Claude Code Ecosystem

- [[anthropics/claude-code#10205](https://github.com/anthropics/claude-code/issues/10205)](https://github.com/anthropics/claude-code/issues/10205) — Stop hook causes infinite loop
[...]

PR #900, opened 2026-07-11
fix: stop-hook no longer blocks turn-ends forever on leaked subagents

### LINKED ISSUE #852: Bug: Stop Hook deadlock after Skill→Agent nested calls — hasRunningSubagents() never cleared
## Description

When Claude Code's built-in `Skill` tool spawns child agents via `Agent(run_in_background: true)`, the Claudian `SubagentManager` tracks these in `activeAsyncSubagents` Map (`main.js` line 94957). However, when agents complete, their completion notifications arrive as `task-notification` system messages destined for the auto-turn callback — but the Stop Hook fires **before** the auto-turn can process them, checks `hasRunningSubagents()` (line 95055), finds the Map still populated, and **blocks** with:

> "Background subagents are still running. Use `TaskOutput task_id="..." block=true` to wait for their results before ending your turn."

This creates a **deadlock**: the auto-turn that would consume the completion notification and clean up the Map can never execute, because the Stop Hook blocks the turn from ending.

## Evidence (main.js, v2.0.21)

**1. Stop Hook blocks when Map is non-empty** (line 95055-95056):
```js
hasRunningSubagents() {
  return this.pendingAsyncSubagents.size > 0 || this.activeAsyncSubagents.size > 0;
}
```

**2. StateProvider feeds into Stop Hook** (line 100610-100613):
```js
tab.service.setSubagentHookProvider(
  () => ({ hasRunning: tab.services.subagentManager.hasRunningSubagents() })
);
```

**3. Cleanup happens in `handleAgentOutputToolResult`** (line 94999-95003), which requires the auto-turn callback to process the `task-notification` — but the Stop Hook fires first:
```js
subagent.asyncStatus = finalStatus;
subagent.status = finalStatus;
subagent.result = extractedResult;
subagent.completedAt = Date.now();
if (agentId) this.activeAsyncSubagents.delete(agentId);
```

**4. Root cause: Stop Hook → auto-turn ordering** (line 68371-68374 + 100610):
```js
hooks.Stop = [createStopSubagentHook(
  () => this._subagentStateProvider?.() ?? { hasRunning: false }
)];
```
The Stop Hook is evaluated before the auto-turn callback (`setAutoTurnCallback`, line 100615) gets a chance to process the `task-notification` system message.

## Repro Steps

1. Use a `Skill` that calls `Agent(run_in_background: true)` (e.g., any hello-ashare skill spawning subagents)
2. Wait for all agents to complete — verified via `TaskList` → empty, `TaskOutput` → all outputs received
3. Try to end the turn → Stop Hook returns `{ decision: "block" }` forever

## Impact

The session enters an **unrecoverable loop** — every turn end is blocked indefinitely, forcing the user to manually close/refresh the tab. `TaskStop` cannot help because there are no actual running tasks. This affects **all** Skill→Agent nested workflows.

## Suggested Fix

**Option A**: In `hasRunningSubagents()`, first try to flush pending auto-turn notifications before checking Map size.

**Option B**: Add an event-loop yield in the Stop Hook to allow the auto-turn callback to fire before the check (e.g., `await new Promise(r => setTimeout(r, 0))`).

**Option C**: Use a separate flag (`_pendingAutoTurnCleanups`) that the auto-turn callback sets, and have `hasRunningSubagents()` also check that flag.

## Environment

- **Claudian version**: 2.0.21
- **Provider**: Claude Code (via Obsidian)
- **OS**: Windows
- **Trigger scenario**: Agent spawned via `Skill` tool (indirect Agent invocation)

Sammoronda:
**Problem** — Turn-ends get blocked forever with "Background subagents are still running" (#852; related history: #493, #513). Once a background subagent dies outside its normal completion path, the Stop hook blocks every subsequent turn-end with no escape — the hook even fails closed when reading state throws.

**Root cause** — Kill/interrupt/stream-error paths never clean `SubagentManager`'s `pendingAsyncSubagents`/`activeAsyncSubagents` maps, so `hasRunningSubagents()` stays true forever. `orphanAllActive()` existed but only ran on tab close / conversation switch / forced new chat.

**Changes**
1. `InputController.sendMessage`: user interrupt (cancel/Esc) and stream errors now orphan active async subagents via the existing `orphanAllActive()` mechanic (approve-new-session excluded — it orphans via `createNew()` already).
2. `createStopSubagentHook`: cap of 3 consecutive Stop blocks, then allow the turn to end with a warning. Covers the catch path (error now logged) and a new state-shape guard (`typeof state?.hasRunning === 'boolean'`). Counter re-arms once state reports no running subagents.
3. `SubagentManager`: `orphanAllActive()`/`markOrphaned()` take an optional reason; new per-agent `orphanByAgentId()` cleans a single agent from all tracking maps.

**No time-based eviction**: legitimately long-running subagents are never evicted; an agent is only orphaned when an actual termination path runs. The Stop-hook cap only ends the block loop, it never touches tracking state.

Tests: 5944 passed incl. 8 new regression tests. Fixes #852

### DIFF
--- a/src/features/chat/controllers/InputController.ts
+++ b/src/features/chat/controllers/InputController.ts
@@ -357,6 +357,7 @@ export class InputController {
     let wasInvalidated = false;
     let didEnqueueToSdk = false;
     let planCompleted = false;
+    let streamErrored = false;
 
     // Lazy initialization: ensure service is ready before first query
     if (this.deps.ensureServiceInitialized) {
@@ -468,6 +469,7 @@ export class InputController {
         );
       }
     } catch (error) {
+      streamErrored = true;
       const errorMsg = error instanceof Error ? error.message : 'Unknown error';
       await streamController.appendText(`\n\n**Error:** ${errorMsg}`);
     } finally {
@@ -486,6 +488,21 @@ export class InputController {
         const didCancelThisTurn = wasInterrupted || state.cancelRequested;
         if (didCancelThisTurn && !state.pendingNewSessionPlan) {
           await streamController.appendText('\n\n<span class="claudian-interrupted">Interrupted</span> <span class="claudian-interrupted-hint">· What should Claudian do instead?</span>');
+        }
+
+        // Termination path for async subagents on interrupt/stream error.
+        // Interrupting the turn (or a query error) kills in-flight background
+        // tasks — without orphaning them here the leaked map entries keep
+        // hasRunningSubagents() true forever and the Stop hook blocks every
+        // subsequent turn-end.
+        // (approve-new-session sets cancelRequested as a control-flow signal;
+        // that path orphans via conversationController.createNew() instead.)
+        if ((didCancelThisTurn && !state.pendingNewSessionPlan) || streamErrored) {
+          this.deps.getSubagentManager().orphanAllActive(
+            streamErrored && !didCancelThisTurn
+              ? 'Stream error ended the turn before the background task completed'
+              : 'Interrupted before the background task completed'
+          );
         }
         streamController.hideThinkingIndicator();
         state.isStreaming = false;
--- a/src/features/chat/services/SubagentManager.ts
+++ b/src/features/chat/services/SubagentManager.ts
@@ -508,17 +508,17 @@ export class SubagentManager {
     this.pendingTasks.clear();
   }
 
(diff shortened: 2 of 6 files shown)


YishenTu:
fixed in #936

PR #902, opened 2026-07-11
Feat: Optional Human-in-the-loop Approval Notification Sound

z3roX882706:
Hi! First-time contributor here - I'm using Claudian daily and
wanted to give something back. This is a small, self-contained feature I've
been running locally for a while and figured others might like it too.

## What it does
Plays an optional sound when a provider asks for approval, so you actually
notice permission prompts when Obsidian isn't the focused window.

- Off by default - opt in via Settings → Human-in-the-loop alerts.
- Ships with a small CC0 default sound so it works the moment you toggle it
  on. Prefer your own? Drop a WAV in `.claudian/sounds/` and it takes over.
- Volume slider + a "Test sound" button. All 10 locales are covered.

## Kept it safe and unobtrusive
- Fire-and-forget at the approval prompt - it never throws and never blocks
  the approval flow.
- The sound file is restricted to a vault-relative `.wav` inside
  `.claudian/sounds/` (no path traversal), checked via RIFF/WAVE magic bytes,
  and capped at 5 MB.
- No `console.*`: failures are silent on the approval path, while the
  "Test sound" button shows a localized Notice instead.

## About the default sound
"Enter Key Press Mechanical Keyboard" by alpinemesh (Freesound), licensed
CC0 1.0 / public domain - free to bundle and redistribute. It's embedded as
base64 (~28 KB) in `src/utils/hitlDefaultSound.ts`, with the source and
license credited in the file header.

## Tests
Added unit tests for path normalization, WAV validation, and the bundled
default. `npm run typecheck && npm run lint && npm run test && npm run build`
all pass locally.

Happy to adjust anything - naming, defaults, the bundled sound, whatever fits
the project best. Thanks for building Claudian!


src/app/settings/defaultSettings.ts
```diff
@@ -2,6 +2,7 @@ import { getDefaultHiddenProviderCommands } from '../../core/providers/commands/
 import { DEFAULT_REASONING_VALUE } from '../../core/providers/reasoning';
 import { type ClaudianSettings } from '../../core/types/settings';
 import { getBuiltInProviderDefaultConfigs } from '../../providers/defaultProviderConfigs';
+import { DEFAULT_HITL_NOTIFICATION_SOUND_PATH } from '../../utils/hitlNotificationSound';
 
 export const DEFAULT_CLAUDIAN_SETTINGS: ClaudianSettings = {
   userName: '',
```
myrenbro9-boop, line 8: UserName:  ' ' ;

PR #911, opened 2026-07-12
Feat (Codex): Render Structured Command Approvals with Readable Labels

z3roX882706:
## Summary

A small, self-contained enhancement to how Codex command-execution approvals
are presented. Codex already sends structured `commandActions` with each
approval request, but they're currently typed loosely (`{ type: string;
command: string }`) and the approval dialog shows only the raw command string.
This adds the missing presentation layer, so the person approving a command
sees what it will actually do in plain language, with the exact command always
shown alongside.

## What changes

**Before:** the approval prompt shows the raw command text.

**After:** the prompt shows labeled actions derived from the structured data,
for example:

- "Read src/main.ts"
- "List files in ./src"
- "Search for TODO in ./src"

with the exact command rendered underneath, unchanged.

**Implementation:**

- Narrow `CommandAction` into a discriminated union that mirrors the real
  payload (`read` / `listFiles` / `search` / `unknown`), carrying the fields
  each variant provides (`name`, `path`, `query`).
- Add `buildCodexCommandApprovalDisplay` plus the `ApprovalCommandDisplay`
  runtime types that turn those actions into readable labels while preserving
  `exactCommand`.
- Pass the built `commandDisplay` through `CodexServerRequestRouter` into the
  approval callback, and render the labeled actions in the InputController
  approval dialog (with styles for the action list).

## Security note (please review this part)

This lives on the human-in-the-loop approval path, which is a sensitive trust
boundary: the labels are built from data the agent supplies, so a crafted
payload could in principle try to make a command read as more benign than it
is. The design mitigates that by **always rendering the exact command
(`exactCommand`) next to the friendly labels**, so the operator's decision is
anchored on the real command, never on the label alone. The `unknown` variant
is a deliberate catch-all: unrecognized action types fall back to the raw
command rather than being dropped or mislabeled.

I'd genuinely appreciate a careful look at this boundary, and I'm happy to
adjust the presentation (for example, making the exact command more prominent
than the labels) if you'd prefer a different emphasis.

## Testing

- `npm run typecheck`: clean.
- Unit tests added for `codexApprovalDisplay` (label formatting per action
  type plus the unknown fallback) and updated `CodexServerRequestRouter` /
  `InputController` tests. The Codex-related suites pass (160 tests).

## Notes

Backward-compatible with the existing consumer (`CodexNotificationRouter` still
reads `commandActions[0].command`). First time contributing to this area, so
happy to tweak naming, labels, or structure to fit the project's conventions.

### DIFF
--- a/src/core/runtime/types.ts
+++ b/src/core/runtime/types.ts
@@ -21,10 +21,22 @@ export interface ApprovalNetworkContext {
   protocol: string;
 }
 
+export interface ApprovalCommandDisplayAction {
+  label?: string;
+  command: string;
+}
+
+export interface ApprovalCommandDisplay {
+  actions: ApprovalCommandDisplayAction[];
+  exactCommand: string;
+}
+
 export interface ApprovalCallbackOptions {
   decisionReason?: string;
   blockedPath?: string;
   agentID?: string;
+  displayToolName?: string;
+  commandDisplay?: ApprovalCommandDisplay;
   decisionOptions?: ApprovalDecisionOption[];
   networkApprovalContext?: ApprovalNetworkContext;
   additionalPermissions?: unknown;
--- a/src/features/chat/controllers/InputController.ts
+++ b/src/features/chat/controllers/InputController.ts
@@ -21,6 +21,7 @@ import {
 } from '../../../core/runtime/QueuedTurn';
 import type {
   ApprovalCallbackOptions,
+  ApprovalCommandDisplay,
   ApprovalDecisionOption,
   ChatRuntimeQueryOptions,
   ChatTurnRequest,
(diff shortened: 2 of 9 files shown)


YishenTu:
Thanks for working on this. After reviewing the implementation, I don’t think the current outcome is meaningful enough to merge.

The structured descriptions only improve the read-only subset represented by `read`, `listFiles`, and `search`. Commands where approval context matters most—writes, deletes, installs, permission changes, Git mutations, and arbitrary scripts—still receive only a generic label or the raw command. This creates an inconsistent experience and gives us limited value for the additional shared runtime and UI complexity. I’m also uncomfortable making the partial interpretation prominent while placing the exact command behind a collapsed section.

I’d be open to revisiting this if a future version can provide materially broader coverage using reliable provider-native metadata, particularly for mutating operations. The exact command should also remain visible and primary, with any interpretation presented as secondary context. We should avoid implementing a local shell parser or heuristic security boundary in Claudian.

The invisible/bidirectional-character warning is independently useful and could be proposed as a smaller, focused change.

If you can produce a broader and more meaningful outcome along these lines, we can discuss it again. For now, I don’t think we should merge the current version.

z3roX882706:
Thanks for the thorough review! Your reasoning makes sense to me.

Agreed on the core point:
The exact command should stay primary and always visible, with any interpretation as secondary context. Inverting that (labels behind the collapse, raw command up front) is the safer default.

I also take the coverage point. The friendly labels only add value on the read-only subset because that is what Codex's commandActions reliably describes today; the mutating operations where approval matters most are exactly where the provider metadata is thinnest. I would rather not paper over that gap with a local parser, which lines up with your no-heuristic-boundary stance. If a future Codex version exposes richer native metadata for mutating commands, I would be glad to revisit a version that meets the broader-coverage bar with the exact command kept primary.

On your suggestion:
I will pull the invisible/control/bidirectional-character warning out into its own small, focused PR, since it stands on its own regardless of the formatting layer. I will close this PR once that one is up.

Thanks again for taking the time.

PR #930, opened 2026-07-14
fix: never lose async subagent completion notifications

SimonSkade1:
**Problem** — A background subagent's completion arrives as a single `system/task_notification` → `async_subagent_result` chunk. If `SubagentManager` misses it, the entry stays tracked as `running` forever, `hasRunningSubagents()` never clears, and the Stop hook blocks every subsequent turn-end with an unfulfillable instruction (`TaskList` is empty and `TaskOutput` reports the task as not found). Observed in the field: one evening session logged 82 consecutive identical stop-blocks. Related: #852, #493.

**Root cause** — Several paths can drop the one-shot completion signal:

1. `handleAsyncSubagentResult()` only did a direct `activeAsyncSubagents.get(agentId)`; a notification carrying the Task tool id (or any id-namespace skew) missed the lookup and was silently dropped.
2. A completion for a task still in `pendingAsyncSubagents` (its Task tool result never yielded an agent id) had no path to clear it at all.
3. An entry already in a terminal state was left in the maps by the `asyncStatus !== 'running'` guard — and `hasRunningSubagents()` counts raw map size.
4. `renderAutoTriggeredTurn()` returned early on a detached tab DOM, dropping the buffered chunks' state along with their rendering.
5. A throw inside the auto-turn render loop lands in `flushAutoTurnBuffer()`'s catch, which shows a Notice and permanently dropped the decrements of all unprocessed chunks.

**Changes**

1. `SubagentManager.handleAsyncSubagentResult()`: resolve the notification id through `taskIdToAgentId` when the direct lookup misses; complete a matching never-promoted pending entry; reconcile already-terminal leftovers out of the maps; sweep `taskIdToAgentId`/`outputToolIdToAgentId` on completion (new `clearAsyncTracking()`).
2. Carry the SDK's `tool_use_id` (`SDKTaskNotificationMessage` provides it) through the `async_subagent_result` chunk and try it during resolution — the exact key the task may have been registered under, instead of relying only on id-match heuristics.
3. `Tab.ts`: new `applyAsyncSubagentResultChunks()` applies completion state transitions without rendering, isolated per chunk. `renderAutoTriggeredTurn()` uses it when the tab DOM is detached, and re-drives unprocessed chunks' state when the render loop throws (idempotent — re-applying a terminal transition is a no-op — then rethrows so the existing Notice still appears).

**Safety** — Strictly notification-driven: entries are only ever touched when the CLI itself declared the task finished, so a genuinely running subagent can never be orphaned by this change. Deliberately no TTL/staleness sweep: there is no liveness oracle over the control protocol, so any TTL would eventually orphan a long-running background agent and silently drop its later result.

Complements (no overlap with) #900, which handles the termination paths (interrupt/stream-error) plus a stop-block cap, and #705, which reconciles the symptom in `hasRunningSubagents()`. This PR fixes the completion path itself so entries don't go stale in the first place.

**Tests** — `SubagentManager`: task-tool-id resolution, never-promoted pending completion, terminal reconciliation, full tracking-map cleanup, unknown-agent no-op. `Tab`: `applyAsyncSubagentResultChunks` forwards only async result chunks. Full suite passes (`typecheck`, `lint`, `test`, `build`).

### DIFF
--- a/src/core/types/chat.ts
+++ b/src/core/types/chat.ts
@@ -156,7 +156,7 @@ export type StreamChunk =
   | { type: 'done' }
   | { type: 'usage'; usage: UsageInfo; sessionId?: string | null }
   | { type: 'context_compacted' }
-  | { type: 'async_subagent_result'; agentId: string; status: 'completed' | 'error'; result?: string }
+  | { type: 'async_subagent_result'; agentId: string; status: 'completed' | 'error'; result?: string; toolUseId?: string }
   | { type: 'subagent_tool_use'; subagentId: string; id: string; name: string; input: Record<string, unknown> }
   | { type: 'subagent_tool_result'; subagentId: string; id: string; content: string; isError?: boolean; toolUseResult?: SDKToolUseResult };
 
--- a/src/features/chat/controllers/StreamController.ts
+++ b/src/features/chat/controllers/StreamController.ts
@@ -1175,7 +1175,8 @@ export class StreamController {
     const handled = this.deps.subagentManager.handleAsyncSubagentResult(
       chunk.agentId,
       chunk.status,
-      chunk.result
+      chunk.result,
+      chunk.toolUseId
     );
 
     await this.hydrateAsyncSubagentToolCalls(handled);
--- a/src/features/chat/services/SubagentManager.ts
+++ b/src/features/chat/services/SubagentManager.ts
@@ -59,6 +59,10 @@ function parseJsonValue(value: string): unknown {
   }
 }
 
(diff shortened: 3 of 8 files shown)


YishenTu:
i have landed my own fix in #936 for the async subagent, would you please test mine and see if the original issue is solved now?

SimonSkade1:
Superseded by #936's native async-subagent lifecycle (v2.0.35): the hook-based tracking this patched was removed entirely, so the completion-loss vector no longer exists. Closing.

PR #956, opened 2026-07-19
fix: wrap input toolbar so permission/mode selector stays visible in narrow sidebars

Iron0utlaw:
## Problem
In a narrow sidebar the input toolbar overflows the pane and clips the
right-aligned selector — hiding the permission mode ("YOLO") on Claude tabs and
the mode selector on Codex/Opencode/Pi tabs. The control becomes unreachable.

Cause: `.claudian-input-toolbar` is a single-line flex row with no `flex-wrap`,
so when its children exceed the pane width, the last item (which has
`margin-left: auto`) is pushed off-screen.

## Fix
Add `flex-wrap: wrap` + `row-gap: 12px` to the toolbar. Auto margins resolve
per-line after wrapping, so the wide layout is unchanged (selector still pinned
right on one row) and in a narrow pane the selector wraps to a second line
instead of disappearing.

## Why not `overflow-x: auto`?
It hides the selector until the user scrolls, and it clips the model/effort
dropdowns — those open upward (`bottom: 100%`), and `overflow-x: auto` computes
`overflow-y` to `auto` too, creating a clip context. `flex-wrap` sets no
`overflow`, so dropdowns are unaffected.

## Tradeoff
In a very narrow pane the toolbar can wrap to two rows — a minor responsive
height change, accepted to keep the control visible.

## Screenshots
Before (clipped):
image

After (wraps, visible):
image

image

## Testing
CSS-only — no unit test to add (per `AGENTS.md`). Verified in Obsidian across
pane widths on both a permission-toggle and a mode-selector tab; dropdowns
unaffected. `lint`, `typecheck`, `build` pass. `styles.css` is gitignored, so
this is a source-only diff.

### DIFF
--- a/src/features/chat/tabs/Tab.ts
+++ b/src/features/chat/tabs/Tab.ts
@@ -1068,6 +1068,8 @@ function initializeInputToolbar(
     },
   });
 
+  dom.eventCleanups.push(() => toolbarComponents.layoutController.destroy());
+
   tab.ui.modelSelector = toolbarComponents.modelSelector;
   tab.ui.modeSelector = toolbarComponents.modeSelector;
   tab.ui.thinkingBudgetSelector = toolbarComponents.thinkingBudgetSelector;
--- a/src/features/chat/ui/InputToolbar.ts
+++ b/src/features/chat/ui/InputToolbar.ts
@@ -17,6 +17,11 @@ import type {
   UsageInfo,
 } from '../../../core/types';
 import { appendCheckIcon, appendMcpIcon, createProviderIconSvg } from '../../../shared/icons';
+import {
+  cancelScheduledAnimationFrame,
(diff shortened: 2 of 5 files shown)


YishenTu:
Follow-up applied in 2fe6e2bb:

- Kept the original flex wrapping behavior as the final fallback.
- Added automatic wrap detection so the toolbar hides the `Effort:` label and context percentage only when the full toolbar would wrap.
- Restores both labels automatically when enough width is available.
- Preserved the context gauge, tooltip, warning state, and added accessible progress semantics.
- Added resize/content observation with explicit cleanup when a tab is destroyed.
- Added focused layout, resize, accessibility, and lifecycle coverage.

Verified locally with typecheck, lint, 6,202 tests, and a production build.

PR #957, opened 2026-07-19
feat: add Claude 5-hour usage guard

Starlight143:
## Summary

- Adds a Usage Guard toggle + threshold slider (50-100%, default 90%) to the Claude settings tab.
- Polls Claude's 5-hour OAuth usage every minute; pauses new sends at the threshold and auto-resumes once usage drops back below it. Fails open on any error.

## Test plan

- `npm run typecheck`, `npm run lint`, `npm test` (new/changed suites) all pass.
- `npm run build` verified via direct esbuild invocation.


Starlight143:
Addressed all three Codex findings in 0b7021f:
- Poll failures after a block now fail open (clear the block) instead of pausing indefinitely.
- createClaudeWorkspaceServices now awaits the first usage check (capped at 5s) before the provider is ready, closing the cold-start race.
- Blocked programmatic resends (e.g. a message popped off the queue) now restore their content to the composer instead of dropping it.

PR #964, opened 2026-07-20
perf: throttle streaming renders, skip hidden/collapsed blocks

SimonSkade1:
## Problem

During streaming, every chunk schedules a `requestAnimationFrame` that re-renders the **entire current block from scratch** in `StreamController` → `MessageRenderer.renderContent`: `el.empty()`, a full `MarkdownRenderer.render`, and re-wrapping every `<pre>`. That runs up to ~60×/sec and is O(block²) over the stream, on the Obsidian render thread. With a few concurrent chat tabs it produces visible UI jank (observed with 3–4 concurrent sessions on a laptop).

Three things compound it:
1. Thinking blocks are collapsed by default, yet the full thinking text is re-rendered every frame into a hidden element.
2. There is no throttle — one full re-render per animation frame.
3. There is no visibility check — a tab/leaf that isn't shown still re-renders every frame.

## Changes

1. **Skip live rendering of collapsed thinking blocks.** The streaming loop no longer renders into a collapsed (hidden) thinking block. `createThinkingBlock` wires an on-expand render of the accumulated content, so expanding mid-stream or after finalize renders the current text once; while expanded it updates live. The deferred-math finalize path only re-renders when the block is expanded.
2. **Throttle streaming text/thinking re-renders** to a minimum interval (150ms) instead of once per frame (~10× less render work, imperceptible). `finalize` forces the final render through the throttle so a completed block is always fully rendered.
3. **Skip rendering while the messages container is hidden** (inactive tab, background leaf, or collapsed sidebar), detected via `offsetParent`. It polls for reveal so a mid-stream switch catches up, and `finalize` renders regardless of visibility.

A small `scheduleDelayedFrame` helper is added to `utils/animationFrame` for the throttle/poll (cancellable via the existing `cancelScheduledAnimationFrame`).

## Notes / scope

- I described the mechanism and the observed jank; I did not run before/after profiling, so the ~10× figure is the reduction in render invocations, not a measured frame-time delta.
- A minor known transient: if a user reveals a background tab during a streaming *pause* (no new chunk in flight), the block updates on the next chunk or at finalize (within the poll interval) rather than instantly.
- Incremental rendering (only re-rendering the trailing incomplete block to kill the O(n²) entirely) is a larger, separate change and is intentionally out of scope here.

## Testing

`npm run typecheck && npm run lint && npm run build` pass. Added/updated unit tests in `StreamController.test.ts` (throttle, hidden-skip + finalize flush, collapsed-thinking skip) and `ThinkingBlockRenderer.test.ts` (on-expand render); the rendering/controller suites pass.

### DIFF
--- a/src/features/chat/controllers/StreamController.ts
+++ b/src/features/chat/controllers/StreamController.ts
@@ -30,6 +30,7 @@ import {
   cancelScheduledAnimationFrame,
   scheduleAnimationFrame,
   type ScheduledAnimationFrame,
+  scheduleDelayedFrame,
 } from '../../../utils/animationFrame';
 import { formatDurationMmSs } from '../../../utils/date';
 import { extractDiffData } from '../../../utils/diff';
@@ -78,6 +79,16 @@ export interface StreamControllerDeps {
   persistConversation?: () => Promise<void>;
 }
 
+/**
+ * Minimum wall-clock gap between successive streaming markdown re-renders of the
+ * same block. Each render redoes the whole block, so rendering every animation
+ * frame is O(block^2). ~150ms is imperceptible while cutting render work ~10x.
+ */
+const STREAMING_RENDER_MIN_INTERVAL_MS = 150;
+
+/** Poll interval for re-checking visibility while a render target is hidden. */
+const HIDDEN_RENDER_POLL_INTERVAL_MS = 250;
+
 export class StreamController {
   private static readonly ASYNC_SUBAGENT_RESULT_RETRY_DELAYS_MS = [200, 600, 1500] as const;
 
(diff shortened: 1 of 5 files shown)


YishenTu:
Thanks for the investigation and for clearly documenting the rendering path. The underlying issue is real: streaming could repeatedly rerender the entire accumulated Markdown block, including hidden or collapsed content, and async renders were not coordinated.

I implemented and merged a replacement in #989. Your analysis directly informed it. The merged version keeps the 150 ms live-render throttle and hidden/collapsed optimizations, with a few changes:

- a centralized latest-wins coordinator allows at most one async render in flight and coalesces intermediate updates
- tab activation and `IntersectionObserver` signals replace visibility polling
- collapsed thinking skips live renders but receives one exact final render, so later expansion is immediate
- forced final renders bypass throttling and visibility gates
- cancellation, visibility catch-up, overlapping renders, thinking expansion, and final math behavior have focused tests

Closing this PR as superseded by the merged implementation. Thanks again for surfacing the performance problem.

Issue #924, opened 2026-07-14
Context meter pins at 100% for 1M-context models — SDK rawMaxTokens (200k) overrides local 1M sizing; no context-1m beta header


newmethods:
### Summary
On 1M-context models (Opus, Sonnet, Fable, Claude 4.6+), the footer context-usage meter shows ~100% and the "approaching limit" warning even when the conversation is far from full. The 1M model option also appears to have effectively gone away in the selector.

### Version
Claudian 2.0.34 (Obsidian desktop).

### Root cause (from reading the bundled source)
`buildUsageInfo` computes:
```
percentage = min(100, round(contextTokens / contextWindow * 100))
```
`contextWindow` prefers `authoritativeContextWindow` (from the embedded Claude Code SDK's `query.getContextUsage().rawMaxTokens`) over the correct local value from `getContextWindowSize()` (`CONTEXT_WINDOW_1M = 1e6`, via `isCurrentOneMillionContextModel`). When the SDK returns the standard **200,000** for a model that actually has a **1,000,000**-token window, the denominator is 5× too small, so the meter pins at 100% and flags `>80%` warning.

Two contributing observations in 2.0.34:
- No `context-1m` beta header appears to be sent anywhere, so the SDK session's context window stays at the standard 200k.
- There is no client-side auto-compaction (only server-sent `compaction` deltas are rendered), so a full-looking meter never triggers any action.

### Expected
For 1M-context models, the meter should divide by 1,000,000 (or the SDK should report the 1M window, presumably gated behind the `context-1m` beta header).

### Actual
Meter shows 100% / "approaching limit" for normal-sized conversations on 1M models; the 1M window is not reflected.

### Repro
1. Select Opus/Sonnet/Fable (a 1M-context model).
2. Have a moderately long conversation (well under 1M tokens, e.g. ~250k).
3. Hover the footer context meter → it reads ~100% with "(Approaching limit…)".

### Notes
Not a data-loss issue — the model still responds coherently; only the denominator/percentage is wrong.

PR #969, opened 2026-07-23
fix(claude): ignore runtime context window that under-reports a 1M model (#924)

newmethods:
## Problem

`resolveContextWindowSize` (`src/providers/claude/types/models.ts`) trusts any positive, finite runtime value from the Claude SDK's `getContextUsage().rawMaxTokens`:

```ts
if (isValidContextLimit(runtimeContextWindow)) {
  return { contextWindow: runtimeContextWindow, source: 'runtime' };
}
```

But that value can come back as the **200k standard window for a 1M-context model**. When it does, the resolver returns `{ contextWindow: 200000, source: 'runtime' }` and never reaches the `model-default` branch that already knows the model is 1M (`getContextWindowSize` → `CONTEXT_WINDOW_1M`).

The result is the footer context meter reading e.g. **`358k / 200k (Approaching limit, run '/compact')`** on a 1M-model session. That reading is self-evidently wrong: 358k tokens cannot fit in a 200k window, so the true window is ~1M and the denominator is under-reported. The percentage and the "approaching limit" warning are both false.

This appears to be the remaining half of #924 — PR #928 added the `resolveContextWindowSize` precedence chain, but the runtime branch still trusts an under-reported window over the model's own known default.

## Fix

Guard the runtime branch so a runtime value can never *shrink* a known-larger model default:

```ts
if (isValidContextLimit(runtimeContextWindow)) {
  const modelDefault = getContextWindowSize(model);
  if (runtimeContextWindow >= modelDefault) {
    return { contextWindow: runtimeContextWindow, source: 'runtime' };
  }
}
return { contextWindow: getContextWindowSize(model), source: 'model-default' };
```

- **Under-report (the bug):** 1M model + runtime 200k → falls through to `model-default` (1M). ✅
- **Legitimate larger gateway window:** runtime ≥ default is still trusted (`source: 'runtime'`), so custom gateways advertising a bigger window keep working. ✅
- **Standard-window models:** haiku + runtime 200k is not an under-report, still trusted. ✅
- **Custom limits:** unchanged, still take precedence over everything. ✅

## Tests

Adds a `resolveContextWindowSize` describe block in `tests/unit/providers/claude/types/types.test.ts` covering: the #924 under-report regression (sonnet/opus/fable), standard-window models, runtime ≥ default, invalid runtime values, no-runtime fallback, and custom-limit precedence.

Verified locally with the full gate: `npm run typecheck && npm run lint && npm run test && npm run build` all pass (full jest suite: 315 suites / 6842 tests green).

## Notes

Display/accounting-only change — it does not alter what the SDK actually does, only which window the meter divides against. I couldn't add a test that reproduces the SDK's raw 200k report directly (that's provider-runtime behavior), so the coverage targets the resolver contract, which is where the decision is made.

### DIFF
--- a/src/providers/claude/types/models.ts
+++ b/src/providers/claude/types/models.ts
@@ -213,7 +213,14 @@ export function resolveContextWindowSize(
   }
 
   if (isValidContextLimit(runtimeContextWindow)) {
-    return { contextWindow: runtimeContextWindow, source: 'runtime' };
+    // The SDK can under-report a 1M-context model's window as the 200k standard
+    // window (issue #924), which pinned the context meter at "approaching limit"
+    // for context that plainly exceeds 200k. A runtime value must never shrink a
+    // known-larger model default; trust it only when it meets or exceeds it.
+    const modelDefault = getContextWindowSize(model);
+    if (runtimeContextWindow >= modelDefault) {
+      return { contextWindow: runtimeContextWindow, source: 'runtime' };
+    }
   }
 
   return {
--- a/tests/unit/providers/claude/types/types.test.ts
+++ b/tests/unit/providers/claude/types/types.test.ts
@@ -700,6 +700,75 @@ describe('types.ts', () => {
     });
   });
 
+  describe('resolveContextWindowSize', () => {
+    it('trusts a runtime window that meets or exceeds the model default', () => {
+      expect(resolveContextWindowSize('sonnet', undefined, CONTEXT_WINDOW_1M)).toEqual({
+        contextWindow: CONTEXT_WINDOW_1M,
+        source: 'runtime',
+      });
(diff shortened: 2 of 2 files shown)


YishenTu:
for custom model, you can specify the context window when you set it up via env var. is it not working?

newmethods:
Ah, thanks for the custom-limit trick. I'm on a custom API proxy whose model IDs (e.g. `claude-sonnet-latest` / `anthropic--claude-4.6-sonnet`) aren't ones Claudian maps to a 1M window, so it defaults to 200k. On standard IDs (`sonnet`, `claude-sonnet-4-6`) Claudian reports 1M. Closing the PR — appreciate the quick look.

Issue #1097, opened 2026-08-13
[Bug]: mermaid only show raw code


xinkyo:
### Before submitting

- [x] I searched existing issues and did not find the same problem.
- [x] I have read the relevant documentation (e.g., the README) but could not resolve this issue.

### Problem

<img width="913" height="508" alt="Image" src="https://github.com/user-attachments/assets/1201df96-d1b9-4561-b1d3-abcb2635bddc" /> mermaid only show code, no graph.

### Steps to reproduce

just ask it to draw mermaid , only code show

### Expected behavior

should be graph instead

### Environment

- Claudian version: most updated
- Obsidian version: most update
- Operating system: win
- Provider: claude
- Provider CLI version: claude code2
- Provider CLI installation method: powershell

xinkyo:
我找到了原因：
Claudian 在聊天消息渲染时，并不是直接把原始 Markdown 交给 Obsidian。 真正的问题

dse(...) 会把所有代码块语言都改成类似这样的占位语言：

claudian-display-only-fence-0
claudian-display-only-fence-1

所以原本：

```mermaid
graph TD
A --> B
```

在进入 Obsidian 的 MarkdownRenderer 之前，已经变成了：

```claudian-display-only-fence-0
graph TD
A --> B
```

这样一来：

Obsidian 看不到 mermaid
就不会触发 Mermaid renderer
渲染结束后 Claudian 再把语言恢复成 mermaid
最终只剩下普通代码块 + Prism 高亮

我改动的具体代码
原始代码

我改的是这段逻辑：

let s=`${vit}${e.length}`,a=o[2];e.push({placeholderLanguage:s,originalLanguage:a});
修改后

我改成：

let s=`${vit}${e.length}`,a=o[2];if(a.toLowerCase()==="mermaid")return i;e.push({placeholderLanguage:s,originalLanguage:a});

xinkyo:
still not fixed

YishenTu:
Opened #1373 to restore Mermaid diagrams in chat.

The implementation uses Obsidian's bundled Mermaid API directly, while keeping all Markdown code-block processors blocked as intended by #1041. Diagrams render after streaming finishes. A source button selects the original code, and leaving the source view returns to the diagram; failed rendering keeps the code available.

This is an independent implementation from #1115. It is not merged or released yet.

PR #1115, opened 2026-08-17
feat: render mermaid diagrams in chat

xiantang:
## Summary

- Add a setting to render Mermaid code fences as diagrams in chat.
- Keep all other code-block processors isolated from assistant content.
- Defer Mermaid rendering until a streamed text block is complete.
- Add localized settings copy, diagram styling, and focused regression coverage.

## Validation

```bash
pnpm run typecheck

pnpm exec jest \
  tests/unit/features/chat/rendering/DisplayOnlyCodeFences.test.ts \
  tests/unit/features/chat/rendering/MessageRenderer.test.ts \
  tests/unit/features/chat/controllers/StreamController.test.ts \
  tests/unit/providers/claude/types/types.test.ts \
  --runInBand

pnpm run lint
```

Lint reports only the repository's existing UI sentence-case warnings.


xiantang:
@YishenTu could u take a look for this PR thanks

xiantang:
@YishenTu  please take a look for this feature thanks

YishenTu:
I found two issues that should be addressed before merge:

1. **Mermaid reopens the generic processor execution boundary by default.** `MessageRenderer` leaves `mermaid` fences unchanged before calling `MarkdownRenderer.render`, so model-generated source can reach any plugin that registers a processor for the `mermaid` language, not only Obsidian's built-in renderer. Because `renderDiagramsInChat` defaults to `true`, existing users automatically lose part of the isolation established by #1041. Please keep all fences neutralized and render Mermaid through a Claudian-owned, narrowly controlled path. If generic processor dispatch is intentional, it should at least be explicit opt-in with the trust tradeoff documented.

2. **Valid Mermaid fences bypass streaming deferral.** `hasDiagramFence` recognizes only an unadorned top-level opener, while the authoritative Markdown segmenter also recognizes blockquoted, list-nested, and attributed fences. Inputs such as `> ```mermaid`, `- ```mermaid`, and ` ```mermaid title="sample"` therefore reach the Mermaid processor on every partial streaming render. Please base detection on `transformMarkdownSegments` rather than a competing regex, and add nested, attributed, and false-positive inner-fence tests.

The allowlist direction is safer than blacklisting Dataview, but these two gaps preserve processor-execution risk and reintroduce the streaming-error behavior for valid Markdown forms.

xiantang:
> I found two issues that should be addressed before merge:
> 
> 1. **Mermaid reopens the generic processor execution boundary by default.** `MessageRenderer` leaves `mermaid` fences unchanged before calling `MarkdownRenderer.render`, so model-generated source can reach any plugin that registers a processor for the `mermaid` language, not only Obsidian's built-in renderer. Because `renderDiagramsInChat` defaults to `true`, existing users automatically lose part of the isolation established by [fix: prevent chat code blocks from running processors #1041](https://github.com/YishenTu/claudian/pull/1041). Please keep all fences neutralized and render Mermaid through a Claudian-owned, narrowly controlled path. If generic processor dispatch is intentional, it should at least be explicit opt-in with the trust tradeoff documented.
> 2. **Valid Mermaid fences bypass streaming deferral.** `hasDiagramFence` recognizes only an unadorned top-level opener, while the authoritative Markdown segmenter also recognizes blockquoted, list-nested, and attributed fences. Inputs such as `` > ```mermaid ``, `` - ```mermaid ``, and ``  ```mermaid title="sample" `` therefore reach the Mermaid processor on every partial streaming render. Please base detection on `transformMarkdownSegments` rather than a competing regex, and add nested, attributed, and false-positive inner-fence tests.
> 
> The allowlist direction is safer than blacklisting Dataview, but these two gaps preserve processor-execution risk and reintroduce the streaming-error behavior for valid Markdown forms.

updated Thanks for commenting

xiantang:
@YishenTu please take a look for the updates thanks

SkyNotSilent (commented):
I found one remaining fallback gap on the current head (`ac03f6c3`).

`renderDiagramBlocks()` hides the neutralized `<pre>` whenever the second `MarkdownRenderer` call resolves, even if the Mermaid processor produces no DOM. The current success test actually exercises that case: the first mock renders the inert source, while the second call is the `beforeEach` resolved no-op. The diagram host therefore has zero children, but the test still expects `claudian-diagram-source-hidden`, leaving the user with an empty block and no readable source.

Could the source be hidden only after the isolated host contains an accepted rendered result? A resolved-but-empty pass should remove the empty host and keep the source visible. I would also change the success fixture so the second call appends a rendered child, then add a separate resolved-no-output regression. That would make the stated failure fallback cover processors that silently return without rendering, not only rejected Promises.

YishenTu:
Reviewed head `ac03f6c3`. The earlier processor opt-in and nested-fence deferral concerns are addressed. Two issues remain:

1. **P2 — Failed rendering hides the source.** In `MessageRenderer.ts:816–818`, a resolved render promise is treated as success. Obsidian catches Mermaid syntax errors internally and resolves after displaying an error; a processor can also resolve without output. Both cases hide the original source and its copy controls. Please confirm successful diagram output before hiding the source, accounting for deferred rendering. This also confirms the empty-output case raised in the existing review. Cover resolved error/empty output, not only rejected promises.

2. **P2 — Mixed-case Mermaid fences do not render.** In `MessageRenderer.ts:797`, the second render preserves the original language casing. Detection accepts `Mermaid` and `MERMAID`, but Obsidian's Mermaid processor selects `code.language-mermaid` case-sensitively. These fences enter the diagram path but remain code. Please pass the canonical lowercase language to the second render and cover a mixed-case fence through `MessageRenderer`.

Validation: all 263 tests across the four focused suites (`DisplayOnlyCodeFences`, `MessageRenderer.diagram.dom`, `MessageRenderer`, and `StreamController`) passed. I also checked the installed Obsidian renderer implementation. No live UI verification.

YishenTu:
Thanks for working on Mermaid support. I opened an independent implementation in #1373 addressing #1097 and #1343.

It calls Obsidian's bundled `loadMermaid().render` API directly, so Mermaid can render by default without passing any fence through registered Markdown code-block processors. It preserves #1041's processor isolation, defers rendering until streaming finishes, and retains source on failure. The source button selects the code and returns to the diagram when focus leaves.

The new PR does not use this branch's implementation. It is still awaiting merge; this PR remains open.

PR #1125, opened 2026-08-20
feat: add native /model and /compact slash commands

SkyNotSilent:
## Summary
- Adds `/model` as a client-side built-in command. With no argument it opens a picker dropdown; with an argument it fuzzy-matches a single model and switches directly. Reuses the existing toolbar model-change path (`ModelSelector.selectValue`) rather than duplicating switching logic.
- Makes `/compact` discoverable in the slash-command dropdown for both providers via a shared `providerRouted` built-in entry. Execution is unchanged — the raw text still reaches the active provider's own compact handling (Claude SDK-native, Codex `thread/compact/start`).
- Removes the Codex-only duplicate `/compact` catalog entry now that it's covered by the shared built-in command list, so Claude and Codex both surface it consistently.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test -- --selectProjects unit` (5585 tests passing, including new coverage for `/model` picking/matching/dismiss flows and updated `/compact` discoverability tests)
- [x] `npm run build`

### DIFF
--- a/src/core/commands/builtInCommands.ts
+++ b/src/core/commands/builtInCommands.ts
@@ -8,7 +8,7 @@
 import { ProviderRegistry } from '../providers/ProviderRegistry';
 import type { ProviderCapabilities, ProviderId } from '../providers/types';
 
-export type BuiltInCommandAction = 'clear' | 'add-dir' | 'resume' | 'fork';
+export type BuiltInCommandAction = 'clear' | 'add-dir' | 'resume' | 'fork' | 'model' | 'compact';
 type BuiltInCommandCapability = 'supportsNativeHistory' | 'supportsFork';
 type BuiltInCommandSupportContext = ProviderId | Pick<ProviderCapabilities, BuiltInCommandCapability>;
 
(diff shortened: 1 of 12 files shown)


SkyNotSilent:
Closing this in favor of #854.

Two reasons:

1. **Scope.** This PR bundled two unrelated things — a `/model` picker with a new dropdown component, and `/compact` discoverability. The `/model` half is a product decision that deserves its own discussion, not a rider on a discovery fix.

2. **Approach.** This PR promoted `/compact` into shared `BUILT_IN_COMMANDS` and deleted Codex's provider-owned `CODEX_COMPACT_COMMAND`. On a closer look that pushes provider-specific knowledge into the shared layer for no gain — the command already exists and is provider-owned, only *discovery* was broken. #854 fixes exactly that by passing the trigger context through the catalog boundary and leaves the provider in charge of its own commands.

#854 has been rebased onto latest `main` and rewritten to the narrower approach: 6 files, +62/-4, no new types, no execution-protocol changes.

`/model` is withdrawn for now. Happy to open a separate issue for it first if there's interest.

Issue #1064, opened 2026-08-06
Claude CLI path field rejects a valid, existing path (Windows, path contains a space)


garciamoises2012:
**Describe the bug**
Setting the "Claude CLI path" field (Claudian settings → Claude tab) to the correct, existing path to claude.exe produces a persistent "Path does not exist" validation error, even though the file demonstrably exists at that exact location.

**To Reproduce**
1. Install Claude Code CLI via the native Windows installer (installs under `%APPDATA%\Claude\claude-code\<version>\claude.exe`)
2. Open Obsidian Settings → Claudian → Claude tab
3. Paste the full path into "Claude CLI path", e.g.:
   `C:\Users\mack mane\AppData\Roaming\Claude\claude-code\2.1.221\claude.exe`
4. Field shows "Path does not exist"

**Verified the path is valid**
- `Test-Path -LiteralPath "<path>"` in PowerShell returns `True`
- `Get-Item -LiteralPath "<path>"` resolves the file successfully (278,279,328 bytes)
- Confirmed byte-for-byte, no stray whitespace/quotes/hidden characters in the pasted string

**Workarounds tried, all fail**
- Windows 8.3 short path to avoid the space in the username folder (`C:\Users\MACKMA~1\AppData\...`) — OS-level `Test-Path` also confirms this resolves correctly, but Claudian still reports "Path does not exist"
- Forward slashes instead of backslashes (`C:/Users/MACKMA~1/AppData/Roaming/Claude/claude-code/2.1.221/claude.exe`) — also rejected
- Ruling out both the space-in-username theory and slash-direction as the cause; the validator appears to reject the path independent of format.

**Suspected related bug**
This may be the same root cause as #1013 ("MCP stdio arguments are incorrectly split when paths contain spaces"), fixed for MCP stdio config on 2026-08-01, if the Claude CLI path field validator lives on a separate code path that fix didn't cover. My Windows username folder itself contains a space ("mack mane"), which lines up with that bug class.

**Environment**
- Claudian version: 2.1.2 (latest)
- Obsidian version: 1.13.4
- OS: Windows 10 Pro
- Claude Code CLI: native installer, version 2.1.221

garciamoises2012:
**Update:** Found a workaround, though the underlying bug is still present.

[...]

PR #1138, opened 2026-08-21
fix: accept a quoted CLI path in provider settings and resolution

SkyNotSilent:
## Why

Related to #1064.

A CLI path pasted with surrounding quotes is rejected by every provider's "CLI path" setting, and the value is then silently discarded. Windows Explorer's **Copy as path** wraps the path in double quotes *exactly when it contains a space*, so this is the normal way a Windows user copies the path they are being asked to paste:

```
"C:\Users\mack mane\AppData\Roaming\Claude\claude-code\2.1.221\claude.exe"
```

What happens today:

1. The validator expands the raw string, so `fs.existsSync('"C:\...exe"')` is `false` → **Path does not exist**.
2. `renderHostnameCliPathSetting` gates persistence on validation (`if (!updateValidation(value)) return;`), so the value is **never saved** — the error is sticky rather than advisory.
3. The resolver has the same gap, so even a saved quoted path would return `null` and fall through to PATH auto-detection.

That produces the confusing combination reported in #1064: the settings field insists the path does not exist, while leaving the field empty and relying on auto-detection works fine.

`stripSurroundingQuotes` already exists in this repo and is already applied to PATH segments (`parsePathEntries`, `parsePathEntriesForPlatform`). The configured-path entry point is the only place it was never applied. This is not a new policy — it closes the one gap in an existing convention.

## What Changed

- `src/utils/path.ts`: exported the existing `stripSurroundingQuotes` and added `normalizeConfiguredCliPath()` = `trim → unquote → expandHomePath`, matching the ordering already used by `parsePathEntries` so `"%APPDATA%\..."` still expands.
- `src/utils/cliBinaryLocator.ts`: dropped the verbatim duplicate of `stripSurroundingQuotes` and imported the shared one; `resolveConfiguredCliPath` now normalizes.
- `src/providers/claude/runtime/ClaudeCliResolver.ts`: `resolveConfiguredPath` now normalizes.
- The five settings validators (claude, codex, grok, opencode, pi) now normalize.

No new types, no change to what is persisted, no change to how the CLI is spawned.

## Why This Approach

**Both ends had to change, not just the validator.** `HostnameCliPathSetting` persists `value.trim()` — the raw string, not a normalized one. Fixing only the validator would let the quoted string through the gate and into settings, where the resolver would still reject it and fall back to PATH. That is strictly worse than today's behavior: the field would look accepted while the configured CLI was quietly ignored. The invariant this PR restores is that **the settings validator and the runtime resolver accept the same set of strings**, and `tests/unit/providers/claude/runtime/ClaudeCliResolver.test.ts` pins it directly.

**Not `expandHomePath` itself.** It also serves PATH segments (`parsePathEntries`) and `normalizePathForFilesystem`; unquoting there would change PATH semantics for unrelated callers.

**Placement inside Grok's validator** is before the `path.posix/win32.isAbsolute` check — otherwise a quoted absolute path fails with "Path must be absolute" instead. Placement inside Codex's validator is after the win32+WSL bypass, leaving that bypass untouched.

Alternative considered and rejected: normalizing inside `HostnameCliPathSetting` so the *stored* value is unquoted. That component is shared and runs `onChange` per keystroke; rewriting the user's text mid-typing is a larger and riskier behavior change than accepting the quoted form at both read points.

## Validation

- `npm run typecheck`
- `npm run lint`
- `npm run test` — 556 suites / 8276 tests
- `npm run build`
- `git diff --check`, clean `git status`

New coverage (no test anywhere previously exercised a quoted or space-containing configured CLI path):

- `tests/unit/utils/path.test.ts` — 8 cases for `normalizeConfiguredCliPath`: double/single quotes, unquoted passthrough, mismatched leading quote left alone, whitespace trimmed before unquoting, env-var and `~` expansion *after* unquoting, blank/undefined input.
- `tests/unit/utils/cliBinaryLocator.test.ts` — resolves a quoted path under a directory containing a space (real temp dir, no fs mock).
- `tests/unit/providers/claude/runtime/ClaudeCliResolver.test.ts` (new) — quoted host-scoped path, quoted legacy path, unquoted path still works, and **`does not fall back to PATH discovery when the configured path is quoted`**, which is the assertion a validator-only fix would fail.
- The five `*SettingsTab` tests — one case each asserting the quoted value is **persisted**, not merely that no error string is rendered; persistence is the actual gate.

## Impact and Follow-ups

- **Tradeoff:** a POSIX filename may legally contain `'` or `"`, so a path that both starts and ends with the same quote character is now unquoted. That is vanishingly rare, and `parsePathEntries` has already made exactly this tradeoff for PATH segments since it was written; this change keeps the two consistent rather than introducing a new rule.
- The stored settings value keeps whatever the user typed, including quotes. Round-tripping is correct because both readers normalize, but the field will display the quoted form. Normalizing at write time is possible as a follow-up if you'd prefer that.
- Nothing changes for unquoted paths — every existing CLI-path test passes unmodified.

Follow-ups I noticed while tracing this and am happy to open separately, deliberately kept out of this PR:

1. Only Grok's validator wraps its `fs` calls in `try/catch`; the other four let a throwing `fs.statSync` (EPERM, ELOOP) escape the Obsidian `onChange` handler.
2. grok/opencode/pi hardcode English `'Path does not exist'` while claude/codex use `settings.cliPath.validation.notExist`.
3. The five validators are near-duplicates with three genuine rule differences (Grok's absolute + `X_OK` checks, Codex's WSL bypass); they could share a base.

Finally, on #1064: the reporter states their pasted string had no quotes, so I am not claiming this fixes their exact input — hence `Related to`, not `Fixes`. It is a reproducible defect in the same code path with the same symptom, and it would be worth having them retest with this change before that issue is closed.

## Checklist

### DIFF
--- a/src/providers/claude/runtime/ClaudeCliResolver.ts
+++ b/src/providers/claude/runtime/ClaudeCliResolver.ts
@@ -3,7 +3,7 @@ import * as fs from 'fs';
 import { getRuntimeEnvironmentText } from '../../../core/providers/providerEnvironment';
 import type { HostnameCliPaths } from '../../../core/types/settings';
 import { getHostnameKey, parseEnvironmentVariables } from '../../../utils/env';
-import { expandHomePath } from '../../../utils/path';
+import { normalizeConfiguredCliPath } from '../../../utils/path';
 import { findClaudeCLIPath } from '../cli/findClaudeCLIPath';
 import { getClaudeProviderSettings } from '../settings';
 
@@ -68,10 +68,9 @@ export class ClaudeCliResolver {
 }
 
 function resolveConfiguredPath(rawPath: string | undefined): string | null {
-  const trimmed = (rawPath ?? '').trim();
-  if (!trimmed) return null;
   try {
-    const expanded = expandHomePath(trimmed);
+    const expanded = normalizeConfiguredCliPath(rawPath);
+    if (!expanded) return null;
     if (fs.existsSync(expanded) && fs.statSync(expanded).isFile()) {
       return expanded;
     }
--- a/src/providers/claude/ui/ClaudeSettingsTab.ts
+++ b/src/providers/claude/ui/ClaudeSettingsTab.ts
@@ -10,7 +10,7 @@ import { renderNativeMcpSettingsSection } from '../../../shared/settings/NativeM
 import { renderProviderEnablementSetting } from '../../../shared/settings/ProviderEnablementSetting';
 import { renderLastEnabledProviderWarning } from '../../../shared/settings/ProviderModelEnablementWarning';
 import { getHostnameKey } from '../../../utils/env';
-import { expandHomePath } from '../../../utils/path';
+import { normalizeConfiguredCliPath } from '../../../utils/path';
 import { getClaudeWorkspaceServices } from '../app/ClaudeWorkspaceServices';
 import {
   getClaudeModelOptions,
@@ -100,7 +100,7 @@ export const claudeSettingsTabRenderer: ProviderSettingsTabRenderer = {
       const trimmed = value.trim();
       if (!trimmed) return null;
 
-      const expandedPath = expandHomePath(trimmed);
+      const expandedPath = normalizeConfiguredCliPath(trimmed);
 
       if (!fs.existsSync(expandedPath)) {
         return t('settings.cliPath.validation.notExist');
--- a/src/providers/codex/runtime/CodexBinaryLocator.ts
+++ b/src/providers/codex/runtime/CodexBinaryLocator.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 
 import { findCliBinaryPath, resolveConfiguredCliPath } from '../../../utils/cliBinaryLocator';
 import { parseEnvironmentVariables } from '../../../utils/env';
-import { expandHomePath } from '../../../utils/path';
+import { expandHomePath, stripSurroundingQuotes } from '../../../utils/path';
 import type { CodexInstallationMethod } from '../settings';
 import type { CodexExecutionTarget } from './codexLaunchTypes';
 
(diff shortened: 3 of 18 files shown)


YishenTu:
Accepting matching surrounding quotes is the right policy here, especially for Windows Explorer Copy as path. I found one remaining validator/runtime mismatch in Codex WSL mode:

- `CodexSettingsTab.validatePath` returns from the WSL branch before `normalizeConfiguredCliPath` runs.
- `resolveCodexCliPath` also only trims the configured WSL command.

As a result, `"/home/user/my tools/codex"` is persisted and forwarded to `wsl.exe` with the quotes still present in the executable argument. A quoted Windows path such as `"C:\Users\user\codex.exe"` also evades `isWindowsStyleCliReference`, so it is accepted instead of producing the intended WSL Windows-path error.

Could we strip matching surrounding quotes before WSL classification and resolution, then add cases for quoted Linux and Windows references? This should use `stripSurroundingQuotes`, not the full `normalizeConfiguredCliPath`, because expanding `~` or environment variables against the Windows host would be incorrect for a Linux-side command.

Aside from that WSL gap, the host-native validator and runtime paths are aligned. The affected local suites pass (8 suites, 165 tests), and the GitHub checks are green.

SkyNotSilent:
Thanks for catching the WSL gap. Pushed 03fa77b6: both `CodexSettingsTab.validatePath` (WSL branch) and the WSL branch of `resolveCodexCliPath` now run `stripSurroundingQuotes` before `isWindowsStyleCliReference` / resolution, so `"/home/user/my tools/codex"` reaches `wsl.exe` unquoted and `"C:\Users\user\codex.exe"` is rejected with the WSL Windows-path error.

Deliberately not using `normalizeConfiguredCliPath` there so `~` / env vars are never expanded against the Windows host for a Linux-side command (covered by a test). Also removed the duplicate local `stripSurroundingQuotes` in `CodexBinaryLocator` in favor of the shared helper. Added quoted Linux / quoted Windows cases at the locator and settings-tab seams.

PR #1145, opened 2026-08-22
fix: emit Codex Responses final answer once

### LINKED ISSUE #1143: [Bug]: Codex (Responses protocol) renders the same final answer twice in one bubble
### Before submitting

- [x] I searched existing issues and did not find the same problem.
- [x] I have read the relevant documentation (e.g., the README) but could not resolve this issue.

## Summary

When using the **Codex provider over the OpenAI **Responses** protocol**, the final assistant answer appears **twice** in a single chat bubble (the exact same text rendered twice, consecutively). When Codex talks to the same endpoint via **chat-completions**, the answer renders only once.

## Environment

- Claudian version: 2.2.3 (also present on 2.2.4)
- Provider: Codex
- Model: Qwen3.6-35B (via a local OpenAI-compatible proxy that supports the Responses API)
- Platform: macOS26.3, Obsidian desktop

## Root cause

Codex emits the **same final answer under two different event representations** over the Responses protocol:

1. `response_item → message → output_text` — the standard Responses-API message content
2. `event_msg → agent_message` — Codex's extended event, carrying the **same full text**

Claudian reconstructs/renders **both**, so one answer is drawn twice. The two strings are byte-for-byte identical.

### Evidence

In the Codex session rollout JSONL (`~/.codex/sessions/.../rollout-*.jsonl`) for a single turn:

```jsonc
// event_msg → agent_message (payload.message)  — 448 chars
{ "type": "event_msg", "payload": { "type": "agent_message",
  "message": "这是一篇，具体内容如下：\n\n## 文档类型\n..." } }

// response_item → message → output_text            — 448 chars
{ "type": "response_item", "payload": { "type": "message", "role": "assistant",
  "content": [ { "type": "output_text", "text": "这是一篇，具体内容如下：\n\n## 文档类型\n..." } ] } }
```

`agent_message.text === output_text.text` → `true` (identical).

This is not the model repeating itself (the `output_text` string itself contains the answer only once) — it is Codex delivering one answer in two event forms, and Claudian rendering both.

## Expected behavior

Only **one** copy of the final answer should be displayed. Suggestions:

- Use `agent_message` for streaming display, then **finalize** the bubble with the `message`/`output_text` (replacing, not appending), **or**
- When appending, **dedupe** against an identical consecutive text (skip `agent_message` if it equals the already-appended `output_text` for the same turn/message).

## Workaround (user-side)

Switching the Codex provider to a **chat-completions** endpoint avoids the duplicate (Codex then emits a single form), but that forgoes the native Responses/agent protocol.

### Steps to reproduce

1. Open Claudian with the **Codex** provider selected.
2. Configure the Codex provider to talk to a **Responses-protocol** endpoint (e.g. a local OpenAI-compatible proxy that serves the Responses API, model `Qwen3.6-35B`).
3. Ask a question that produces a normal text answer (e.g. “这是一篇什么文档” on a linked note).
4. Wait for the turn to complete.
5. Observe the final answer bubble: the **same full text is rendered twice, consecutively**, in the one bubble.
6. (Control) Repeat with the same endpoint but a **chat-completions** protocol: the answer renders **once**. This confirms the duplication is specific to the Responses protocol.

### Expected behavior

- The final assistant answer should be displayed **once** per turn, regardless of protocol.
- Because Codex delivers the same answer in two event forms over the Responses protocol — `event_msg → agent_message` and `response_item → message → output_text`, byte-for-byte identical — Claudian should render **only one** of them (e.g. use `agent_message` for streaming, then finalize with `output_text`/`message`, replacing instead of appending; or dedupe identical consecutive text for the same turn/message).

### Environment

- Claudian version: **2.2.3** (also still present in **2.2.4**)
- Obsidian version: desktop, latest
- Platform: macOS
- Provider: **OMLX open-ai compliant api**
- Model: `Qwen3.6-35B-A3B-OptiQ-4bit` via local OpenAI-compatible proxy (Responses API)
- Note: duplication does **not** occur when the Codex provider is pointed at a chat-completions endpoint

### Logs and other evidence

_No response_

### Additional context

_No response_

SkyNotSilent:
## Why

Fixes #1143.

Codex Responses-compatible runtimes can expose the same final assistant text through fallback notifications before the canonical `agentMessage` item ID is available. `CodexNotificationRouter` retained that text at the current anonymous-segment level, while canonical completion deduplication consulted the item-scoped accumulator. When the item ID arrived late, the empty item-scoped state caused the full answer to be emitted again in the same bubble.

This affects live rendering only. The persisted JSONL replay path already deduplicates adjacent `event_msg` and `response_item` representations.

## What Changed

- Seed the item-scoped assistant-text accumulator when an existing anonymous segment first claims its canonical item ID.
- Preserve any text already tracked for that item ID instead of overwriting it with anonymous segment state.
- Add regression coverage for reported Responses-style fallback ordering, recovery without `item/started`, prefix completion, non-prefix completion, and late completion of an already tracked item.

## Why This Approach

`CodexNotificationRouter` owns projection of provider-native live notifications into stream chunks, so reconciling its two text accumulators at the point where an anonymous segment gains an ID keeps the behavior provider-local.

The fix does not add text replacement semantics to shared stream chunks, perform chat-layer content deduplication, or assume that raw and canonical item IDs match. Those alternatives would either leak Codex protocol behavior into shared layers or be fragile with older and compatible runtimes. Non-prefix completions are still emitted in full, and the existing per-item state wins when it is already present.

## Validation

- TDD red: the Responses fallback, recovery-without-start, and prefix-completion cases failed before the implementation by emitting duplicate/full text.
- `npm run test:unit -- --runTestsByPath tests/unit/providers/codex/runtime/CodexNotificationRouter.test.ts --runInBand` — 1 suite / 134 tests passed.
- `npm run typecheck`
- `npm run lint`
- `npm run test` — 565 suites / 8,334 tests passed, plus 14 protocol suites / 131 tests and 61 architecture checks.
- `npm run build`
- `git diff --check`

## Impact and Follow-ups

The change is limited to live Codex notification projection. It does not modify transcript history, storage, raw item ID handling, public types, provider interfaces, or shared execution protocols. No user-facing documentation change is needed.

If #1143's reporter confirms that duplication also survives reopening the conversation, the wider sanitized JSONL record window should be investigated separately rather than expanding this live-path fix.

## Checklist

### DIFF
--- a/src/providers/codex/runtime/CodexNotificationRouter.ts
+++ b/src/providers/codex/runtime/CodexNotificationRouter.ts
@@ -189,6 +189,15 @@ export class CodexNotificationRouter {
 
     if (!this.currentAssistantSegmentId) {
       this.currentAssistantSegmentId = itemId;
+      if (
+        this.currentAssistantSegmentText
+        && !this.streamedAgentMessageTextById.has(itemId)
+      ) {
+        this.streamedAgentMessageTextById.set(
+          itemId,
+          this.currentAssistantSegmentText,
+        );
+      }
     }
   }
 
--- a/tests/unit/providers/codex/runtime/CodexNotificationRouter.test.ts
+++ b/tests/unit/providers/codex/runtime/CodexNotificationRouter.test.ts
@@ -17,6 +17,53 @@ describe('CodexNotificationRouter', () => {
   });
 
   describe('text streaming', () => {
+    const emitEventMessage = (text: string) => {
+      router.handleNotification('event_msg', {
+        type: 'agent_message',
+        message: text,
+      });
+    };
+
+    const emitRawAssistantMessage = (text: string) => {
+      router.handleNotification('rawResponseItem/completed', {
+        threadId: 't1',
+        turnId: 'turn1',
+        item: {
+          type: 'message',
+          role: 'assistant',
+          content: [{ type: 'output_text', text }],
+        },
+      });
+    };
+
+    const startAgentMessage = (itemId: string) => {
+      router.handleNotification('item/started', {
+        threadId: 't1',
+        turnId: 'turn1',
+        item: {
+          type: 'agentMessage',
+          id: itemId,
+          text: '',
+          phase: 'streaming',
+          memoryCitation: null,
+        },
+      });
+    };
+
+    const completeAgentMessage = (itemId: string, text: string) => {
+      router.handleNotification('item/completed', {
+        threadId: 't1',
+        turnId: 'turn1',
+        item: {
+          type: 'agentMessage',
+          id: itemId,
+          text,
+          phase: 'final',
+          memoryCitation: null,
+        },
+      });
+    };
+
     it('maps item/agentMessage/delta to a text chunk', () => {
       router.handleNotification('item/agentMessage/delta', {
         threadId: 't1',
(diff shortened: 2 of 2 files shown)


YishenTu:
I can reproduce the duplicate only by manually feeding `CodexNotificationRouter` the notification order introduced by this PR's tests. That proves the router has an edge case, but it does not yet establish that #1143 reaches that path.

The evidence in #1143 is persisted JSONL. We already cover the same adjacent `event_msg -> agent_message` / `response_item -> assistant message` pair in `CodexHistoryStore.test.ts`, where replay produces one assistant message. For live output, normal Codex 0.149 app-server handling emits the canonical item completion before the raw response item is recorded/emitted, so the existing item-scoped accumulator is already populated. The standard Claudian Codex path also has not reproduced the duplicate.

A late item ID is plausible through Claudian's missed-completion recovery, but the issue currently has no `codex --version` or live notification trace showing that the reporter missed the canonical lifecycle and entered recovery. Could we get a sanitized method-only trace containing `item/started`, `item/agentMessage/delta`, `item/completed`, `rawResponseItem/completed`, `event_msg`, `thread/status/changed`, and `turn/completed`?

There is also an untested message-boundary risk. If anonymous fallback text is emitted after an earlier assistant segment, a later `assistant_message_start` opens a new UI bubble. Seeding the per-item accumulator then suppresses the completion text for that new bubble, potentially leaving it empty and attaching the answer to the previous bubble. The router tests assert filtered text chunks only, so they do not cover the final message structure.

Until the real live order is captured and covered through the execution/UI boundary, I don't think this is validated as `Fixes #1143`. It may still be useful defensive hardening, but the current evidence does not show that it fixes the reported setup.

SkyNotSilent:
Thanks — the message-boundary concern reproduces.

I added a local seam test that wires `CodexNotificationRouter` through `adaptCodexStreamChunk` into `InputController` and asserts the final assistant messages rather than filtered text chunks. For the sequence with an earlier assistant segment, a tool boundary, anonymous fallback text, and a late canonical item:

- with this PR's seeding change, the final assistant contents are `["FirstAnswer", ""]`;
- without the seeding change, they are `["FirstAnswer", "Answer"]`.

So the current change trades the duplicate for an empty canonical bubble while still attaching fallback text to the previous bubble. Neither result satisfies the message-boundary contract, and #1143 still has no live method-order trace establishing that the reporter reaches this path.

I'm withdrawing this PR rather than reframing an unsafe change as defensive hardening. If a live trace becomes available, a future fix should establish the correct assistant boundary before projecting fallback text and cover the final execution/UI message structure. Thanks for catching this.

PR #1146, opened 2026-08-22
feat: surface Claude prompt suggestions in chat

### LINKED ISSUE #1136: [Feature]: Surface prompt suggestions in the chat panel (SDK already supports it)
### Before submitting

- [x] I searched existing issues and did not find the same request.

### Problem or unmet use case

Claudian does not currently implement the `promptSuggestionEnabled` config setting. Because of this, users have no way to see suggested follow-up prompts in the chat panel.

The bundled Claude Agent SDK supports prompt suggestions (its options schema includes `promptSuggestionEnabled`, described as "When false, prompt suggestions are disabled. When absent or true, prompt suggestions are enabled."), but Claudian never renders them, so the capability is inert.

This affects anyone using Claudian for exploratory or multi-step vault work. Suggested prompts increase efficiency by decreasing the amount of typing required and also aid in discovery of useful next steps — after a note analysis, natural follow-ups like "_add these tags_," "_create wikilinks to related notes_," or "_split this into atomic notes_" would otherwise go unnoticed by users who don't already know what to ask. It particularly helps users who rely on the tool for executive-function support, where reducing the "what do I do next?" decision cost matters.

### Desired outcome

Claudian mirrors Claude Desktop's `promptSuggestion` behavior. After an assistant response, a contextually relevant suggested prompt is visible in the chat panel, and using right-arrow or tab (standard conventions) pre-fills it as the next message. Users who find suggestions noisy can turn them off.

### Affected provider

Claude

### Proposed solution

Mirror Claude Desktop behavior by rendering the SDK-provided prompt suggestions as grayed text in the chat box. Only applies if user's config settings include `promptSuggestionEnabled=true`

Optional: Gate by a plugin setting or in-chat toggle that maps to the SDK's `promptSuggestionEnabled` option. Tradeoffs: Suggestion generation may add token cost — another reason to keep it toggleable.

### Alternatives considered

Today the closest workaround is slash commands as user-authored prompt templates. That works for known, repeated prompts but is static: it can't surface context-aware follow-ups the user didn't think to pre-author, and it requires manual setup per vault or user scope.

### Additional context

Verified against Claudian 2.1.4: `promptSuggestionEnabled` appears exactly once in `main.js` (the bundled SDK's schema definition) and `promptSuggestions` twice (init-config passthrough only); no code path displays suggestions in the UI. Environment: Obsidian on macOS, Claude Code backend.

SkyNotSilent:
## Why

Claudian already bundles an Agent SDK version that can generate contextual follow-up prompts, but main-chat queries never enable the capability and the trailing `prompt_suggestion` message is not represented across the execution boundary. Users therefore cannot see or accept the SDK suggestion described in #1136.

Fixes #1136

## What Changed

- Enable `promptSuggestions` only for persistent Claude main-chat sessions; ephemeral title, inline-edit, and auxiliary runs remain unchanged.
- Normalize a non-empty native suggestion into a provider-neutral session event associated with the completed requested turn.
- Preserve the completed turn token after `result`, including second and later turns, while rejecting stale, superseded, and provider-triggered background suggestions.
- Show the suggestion as transient muted placeholder text only in the empty composer of the active tab.
- Accept it with unmodified Tab or Right Arrow, after existing mode and composer-dropdown keyboard handlers.
- Clear it on typing, tab/conversation/provider/model/context changes, instruction or bang-bash mode, and teardown. Suggestions are never persisted.

## Why This Approach

Generation and opt-out policy stay native to Claude: the SDK still respects `promptSuggestionEnabled: false`, the environment override, plan mode, API errors, and its first-turn suppression. Claudian only adapts the resulting UI intent.

A session-scoped event is necessary because the SDK emits the suggestion after the requested event stream has already ended at `result`. Carrying the originating turn ID and the existing session binding fences avoids reopening the completed stream or storing UI-only data in provider state. The composer uses its existing placeholder styling instead of adding a second overlay DOM surface, and the controller revision prevents a delayed event from reviving stale text.

## Validation

TDD red/green coverage includes:

- persistent enablement and ephemeral exclusion;
- second-turn post-result correlation;
- provider-triggered background isolation;
- whitespace filtering;
- Tab and Right Arrow acceptance plus modifier/IME guards;
- active-turn settling, inactive-tab rejection, stale-composer fencing, and dropdown precedence.

Full validation:

- `npm run typecheck`
- `npm run lint`
- `npm run test` — 566 suites / 8,340 tests, plus 14 protocol suites / 131 tests and 61 architecture checks
- `npm run build`
- `git diff --check`

## Impact and Follow-ups

No persisted schema, provider state, transcript, setting, or auxiliary execution behavior changes. Suggestions remain subject to the native SDK suppression rules and are intentionally discarded rather than restored across tab or conversation changes. No documentation update is needed for this focused UI behavior.

## Checklist

### DIFF
--- a/src/core/execution/ProviderExecutionEvent.ts
+++ b/src/core/execution/ProviderExecutionEvent.ts
@@ -305,6 +305,14 @@ export type ProviderSessionErrorEvent = ProviderEventBase<
     readonly recoverable: boolean;
   };
 
+export type ProviderPromptSuggestionEvent = ProviderEventBase<
+  'prompt_suggestion',
+  ProviderSessionEventScope
+> & {
+  readonly originatingTurnId: string;
+  readonly suggestion: string;
+};
+
 export type ProviderBackgroundOutputEvent =
   | (ProviderUserMessageStartedEvent & { readonly scope: ProviderBackgroundEventScope })
   | (ProviderAssistantMessageStartedEvent & { readonly scope: ProviderBackgroundEventScope })
@@ -325,6 +333,7 @@ export type ProviderSessionEvent =
   | ProviderBackgroundOutputEvent
   | ProviderBackgroundTurnCompletedEvent
   | ProviderAsyncSubagentCompletedEvent
+  | ProviderPromptSuggestionEvent
   | (ProviderSessionStateChangedEvent & { readonly scope: ProviderSessionEventScope })
   | (ProviderModeChangedEvent & { readonly scope: ProviderSessionEventScope })
   | ProviderSessionErrorEvent;
--- a/src/core/execution/index.ts
+++ b/src/core/execution/index.ts
@@ -21,6 +21,7 @@ export {
   type ProviderExecutionEventScope,
   type ProviderModeChangedEvent,
   type ProviderNoticeEvent,
+  type ProviderPromptSuggestionEvent,
   type ProviderRequestedEventScope,
   type ProviderRequestedExecutionEvent,
   type ProviderSessionErrorEvent,
--- a/src/features/chat/tabs/TabLifecycle.ts
+++ b/src/features/chat/tabs/TabLifecycle.ts
@@ -50,6 +50,7 @@ export function activateTab(tab: AssembledTabRuntime): void {
 }
 
 export function deactivateTab(tab: AssembledTabRuntime): void {
+  tab.ui.promptSuggestionController.clear();
   tab.controllers.streamController.setTabActive(false);
   tab.dom.contentEl.addClass('claudian-hidden');
   tab.controllers.selectionController.stop();
--- a/src/features/chat/tabs/TabSessionEvents.ts
+++ b/src/features/chat/tabs/TabSessionEvents.ts
@@ -70,6 +70,29 @@ async function handleTabSessionEvent(
     new Notice(event.message);
     return;
   }
+  if (event.type === 'prompt_suggestion') {
+    if (
+      tab.dom.contentEl.hasClass('claudian-hidden')
+      || tab.lifecycleState === 'closing'
+      || tab.ui.instructionModeManager.isActive()
+      || tab.ui.bangBashModeManager?.isActive()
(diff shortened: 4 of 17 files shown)


SkyNotSilent:
#1146 is now conflicting with current `main`. Before I refresh it, could you confirm the intended product boundary?

The current version enables Claude's native suggestions by default for persistent chats, relies on Claude's own `promptSuggestionEnabled: false` as the opt-out, shows one transient suggestion only in an empty composer, and accepts it with unmodified Tab or Right Arrow.

If a discoverable Claudian control is preferred, I recommend a Claude-specific default-off toggle because generation may add provider cost/latency; other providers would remain unchanged. Please also confirm whether the ghost-text plus Tab/Right Arrow interaction is desired, or whether a visible action/chip is preferred.

I'll leave the code unchanged until that direction is clear.

YishenTu:
the main concern is the necessity of this feature and the potential tradeoff (e.g. token consumption)

SkyNotSilent:
Thanks — I audited the exact SDK/CLI path rather than relying on green CI. Enabling `promptSuggestions` starts a separate auxiliary model request after an eligible completed turn. The SDK allows at most one per turn, suppresses it on the first turn, after API errors, and in plan mode, and reuses the parent prompt cache, which Anthropic describes as “nearly free.”

That is not the same as zero cost. The emitted suggestion event contains no suggestion-specific usage, cost, cache, or latency data, so Claudian cannot quantify the tradeoff without controlled authenticated requests. The current PR also opts every persistent Claude chat into generation; opt-out exists only through Claude’s native setting or environment variable, not through a discoverable default-off Claudian control.

#1136 currently provides one user request but no broader validation that justifies making this default-on or carrying the additional cross-layer lifecycle. Given that evidence, I don’t think the current PR clears the necessity/cost bar. I’m going to close it and leave #1136 open for future product direction.

If prompt suggestions are wanted later, I recommend revisiting them as an explicit Claude-only, default-off opt-in and rebuilding the implementation on current `main`.

Issue #729, opened 2026-06-02
Response duration ("* Baked for 7s") not persisted across conversation reloads


gugu123a:
**Describe the bug**

The response footer flavor text (e.g., `* Baked for 7s`, `* Crunched for 12s`) displays correctly when a response first completes. However, after closing and reopening the conversation, the duration seconds are lost — only the flavor word remains (e.g., `* Baked`), and thinking blocks show "Thought" instead of "Thought for Xs".

**To Reproduce**

1. Send a message and wait for the response
2. See `* Baked for 7s` in the response footer (correct)
3. Close the conversation tab
4. Reopen the conversation
5. The footer now shows `* Baked` without the seconds

**Expected behavior**

The duration should persist across conversation reloads so users can see how long each response took.

**Technical notes**

The `durationSeconds` and `durationFlavorWord` fields are set on the assistant message object at the end of streaming (around line 90321 in the built main.js). However, these custom fields appear to be stripped during the SDK save/load cycle — when the conversation is reloaded, `buildAssistantMessage` recalculates `durationSeconds` from API timestamps (which may return null or inaccurate values), resulting in `void 0`. The flavor word is also affected since it's not included in `buildAssistantMessage`'s return value.

**Suggested fix**

Include `durationSeconds` and `durationFlavorWord` in the `buildAssistantMessage` return object, and check for stored values before falling back to API timestamp calculation.

**Environment**

- Claudian: 2.0.21
- Obsidian: desktop
- OS: Windows

PR #1149, opened 2026-08-22
fix: restore Claude response duration from history

SkyNotSilent:
## Why

Refs #729.

Claude response duration is currently added only to the in-memory chat message when a live response finishes, so the footer disappears after the conversation is reloaded. Claude transcripts already persist authoritative system/turn_duration records, but the history loader did not consume them.

## What Changed

- Decode valid Claude turn_duration metadata and correlate it to the final native assistant entry through parentUuid.
- Restore durationSeconds on the merged assistant message while preserving the existing suppression for interrupted turns and sub-second responses.
- Add regression coverage for multi-entry tool-use turns, sub-second metadata, missing metadata, and interrupted turns.

## Why This Approach

The history boundary now consumes provider-native, read-only timing metadata instead of estimating duration from message timestamps or introducing a second persistence format. This keeps live execution unchanged, preserves the existing merged-message model, and leaves older transcripts without timing metadata untouched.

## Validation

- npm run test:unit -- --runTestsByPath tests/unit/utils/sdkSession.test.ts --runInBand (165 tests passed)
- npm run test -- --testTimeout=15000 (573 suites / 8404 tests plus 14 protocol suites / 131 tests passed; the wider timeout avoids a local real-server Collab timing flake that passes in isolation)
- npm run typecheck
- npm run lint
- npm run build
- git diff --check

## Impact and Follow-ups

This affects only Claude history replay. Live streaming and transcript files are unchanged. Older transcripts without turn_duration records continue to omit the footer, and durations below one second remain hidden to match live behavior.

Claude does not persist Claudian's randomly selected footer flavor word or per-thinking-block elapsed time. Reloaded responses therefore use the renderer's existing default flavor, and thinking-block duration reconstruction remains out of scope.

## Checklist

### DIFF
--- a/src/providers/claude/history/ClaudeHistoryStore.ts
+++ b/src/providers/claude/history/ClaudeHistoryStore.ts
@@ -2,7 +2,12 @@ import * as fs from 'node:fs/promises';
 import * as path from 'node:path';
 
 import type { ProviderHistoryPathContext } from '../../../core/providers/types';
-import type { ChatMessage, SubagentInfo, ToolCallInfo } from '../../../core/types';
+import {
+  type ChatMessage,
+  isCanonicalUserMessage,
+  type SubagentInfo,
+  type ToolCallInfo,
+} from '../../../core/types';
 import { ClaudeTaskToolNormalizer } from '../normalization/ClaudeTaskToolNormalizer';
 import { isClaudeSubagentToolName } from '../subagentToolNames';
 import { buildAsyncSubagentInfo } from './sdkAsyncSubagent';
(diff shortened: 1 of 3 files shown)


YishenTu:
Review findings

The Claude history boundary is the right owner for native timing metadata, but two transcript-correlation paths still prevent the stated behavior.

1. **[P1] Resolve turn durations through intermediary system records** — [ClaudeHistoryStore.ts:264](https://github.com/YishenTu/claudian/blob/f4a6b95f34592a879856c30a574ea33c0d58edaa/src/providers/claude/history/ClaudeHistoryStore.ts#L264-L280)

   collectNativeTurnDurations indexes timing by the immediate parentUuid of turn_duration, while flushPendingAssistant looks it up using the final assistant UUID. With Stop hooks, real Claude transcripts use assistant A → stop_hook_summary S → turn_duration(parentUuid=S). The map therefore contains S, lookup requests A, and duration restoration silently fails for those sessions. A captured transcript demonstrates the exact chain: https://gist.github.com/phact/442c90987ba7736c4482464485209730

   Please add a regression fixture with stop_hook_summary between the assistant and turn_duration, and resolve ancestry through non-message system records to the nearest assistant entry.

2. **[P2] Preserve checkpoint timing metadata before resumeAtMessageId truncation** — [ClaudeHistoryStore.ts:149](https://github.com/YishenTu/claudian/blob/f4a6b95f34592a879856c30a574ea33c0d58edaa/src/providers/claude/history/ClaudeHistoryStore.ts#L149-L154)

   filterActiveBranch truncates at the selected assistant UUID before durations are collected. Since turn_duration is a child of that checkpoint, fork and rewind hydration drops the metadata and the checkpoint response still loses its footer after reload. Collect timing ancestry from the untruncated transcript, then apply it only to assistant entries retained by branch filtering. A regression should cover assistant → turn_duration → later turn loaded with the assistant UUID as resumeAtMessageId.

The existing tests and CI are green, but the new fixtures cover only direct-parent, non-checkpoint transcripts.

SkyNotSilent:
Thanks for the concrete transcript and checkpoint cases. Addressed both findings in `78194b78`.

`collectNativeTurnDurations` now builds its index from the untruncated native transcript and resolves a duration parent through consecutive system records to the nearest assistant. Branch filtering still owns which assistant entries are retained, so timing from discarded branches is never rendered.

Regression coverage now includes `assistant -> stop_hook_summary -> turn_duration` and `assistant -> turn_duration -> later turn` loaded with the assistant as `resumeAtMessageId`.

Validation: focused history tests pass 167/167; the full suite passes 573 suites / 8,406 tests, architecture checks pass 61/61, and typecheck, lint, build, and diff checks pass.

Issue #580, opened 2026-04-29
win11，一在obsidian中打开claudian时便卡顿，在缩放claudian插件页面时卡顿尤其明显，obsidian直接卡死


LinkLeft:
win11，一在obsidian中打开claudian时便卡顿，在缩放claudian插件页面时卡顿尤其明显，obsidian直接卡死

yummy527:
> win11，一在obsidian中打开claudian时便卡顿，在缩放claudian插件页面时卡顿尤其明显，obsidian直接卡死

I also encountered this stuck but on linux. Not sure if this is related to large numbers of markdown notes. In my vault, there's a total of more than 10 thousands notes. What about yours?

isolature:
是的，我也是win11,电脑是华硕天选6pro,也是在ob中一使用claudian中整个ob便卡住重启了

LinkLeft:
@yummy527 我的库里就十几个笔记，打开的笔记也只有几百行，即使没有笔记处于预览状态打开claudian也会卡死

skylar010303:
我也遇到了同样的问题

AAA213:
我也遇到相同的问题  claudian 一个对话过长就会出现 新开一个对话就不卡顿了 猜测是对话太多 claudian  插件渲染了所有对话 导致的卡顿

EmailCai:
我现在用的也卡了，打几个字都卡，能解决下吗

STJWLS:
> 我也遇到相同的问题 claudian 一个对话过长就会出现 新开一个对话就不卡顿了 猜测是对话太多 claudian 插件渲染了所有对话 导致的卡顿

我也遇到了这个问题

SkyNotSilent:
I isolated one concrete cost that matches the reports where lag grows with conversation length: every composer-height or sidebar-width change currently reflows all rendered messages, including the offscreen transcript.

PR #1155 applies browser-native offscreen layout isolation per message, with separate intrinsic estimates for short user messages and longer assistant messages. In a Chromium 151 benchmark with 100 alternating messages and 100 forced width/composer-height changes, the measured layout loop dropped from 77.5–89.3 ms to 5.1–6.3 ms (about 93%), while the initial aggregate scroll estimate stayed within 0.37% of the fully laid-out baseline. The full repository CI is green.

[...]

PR #1155, opened 2026-08-22
perf: isolate offscreen chat message layout

SkyNotSilent:
## Why

Long conversations make every composer-height or sidebar-width change reflow the full rendered transcript, including hundreds of messages far outside the viewport. This matches the long-conversation typing and resize jank reported in #580.

## What Changed

- Apply `content-visibility: auto` to each rendered chat message so Chromium can skip layout and paint work for offscreen transcript content.
- Prevent skipped flex children from shrinking when their contents are not laid out.
- Use role-shaped intrinsic fallbacks: `5rem` for typically short user messages and `23.5rem` for typically longer assistant messages.
- Add a style regression test that pins the rendering isolation and both fallback sizes.

Refs #580

## Why This Approach

This uses the same browser-native rendering primitive already used by Claudian's long Collab review/conflict surfaces. It leaves message creation, streaming updates, history loading, scrolling state, selection, and provider behavior unchanged; visible messages continue to render normally, while `auto` allows Chromium to retain a message's measured size after it has been rendered.

The role-specific fallbacks avoid treating a short prompt and a long assistant response as the same initial scroll estimate. `flex-shrink: 0` is necessary because skipped flex children otherwise lose their content-derived minimum size and collapse.

## Validation

- Reproducible Chromium 151 benchmark in an ignored `.context/` fixture: 100 alternating short-user/long-assistant messages and 100 forced sidebar-width/composer-height changes.
  - Run 1: `77.5 ms` baseline vs `5.1 ms` isolated (`93.4%` reduction).
  - Run 2: `89.3 ms` baseline vs `6.3 ms` isolated (`92.9%` reduction).
  - Aggregate scroll-geometry checksum: `2,700,000` baseline vs `2,690,000` isolated (`0.37%` difference before actual sizes are learned).
- TDD red: the focused style test failed because messages had neither flex preservation nor offscreen layout isolation.
- `npm run test:unit -- --runTestsByPath tests/unit/style/components/messages.test.ts --runInBand`
- `npm run typecheck`
- `npm run lint`
- `npm run test` (14 protocol suites / 131 tests, 574 main suites / 8401 tests, 61 architecture and policy tests)
- `npm run build`
- `git diff --check`

## Impact and Follow-ups

This PR addresses the transcript-layout component of #580. It deliberately uses `Refs` rather than `Fixes` because that issue also contains older reports without a confirmed long-conversation trigger; those may have different causes.

No persistence, protocol, provider, or data compatibility impact. Browsers without `content-visibility` support ignore the optimization and retain the current behavior.

## Checklist

### DIFF
--- a/src/style/components/messages.css
+++ b/src/style/components/messages.css
@@ -284,8 +284,6 @@
 }
 
 .claudian-message {
-  flex-shrink: 0;
-  content-visibility: auto;
   padding: 10px 14px;
   border-radius: 8px;
   max-width: 95%;
@@ -293,7 +291,6 @@
 }
 
 .claudian-message-user {
-  contain-intrinsic-size: auto 5rem;
   position: relative;
   background: rgba(0, 0, 0, 0.3);
   align-self: flex-end;
@@ -311,6 +308,8 @@
 }
 
 .claudian-message-assistant {
+  flex-shrink: 0;
+  content-visibility: auto;
   contain-intrinsic-size: auto 23.5rem;
   background: transparent;
   align-self: stretch;
--- a/src/style/components/messages.css
+++ b/src/style/components/messages.css
@@ -284,6 +284,8 @@
 }
 
 .claudian-message {
+  flex-shrink: 0;
+  content-visibility: auto;
   padding: 10px 14px;
   border-radius: 8px;
   max-width: 95%;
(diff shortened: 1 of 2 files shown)


YishenTu:
Added follow-up commit 7b2c843f7 to preserve user-message actions. content-visibility introduces paint containment, while the copy/rewind/fork toolbar is positioned below the user bubble, so applying containment to every message clipped those controls. The repair scopes flex preservation, offscreen rendering, and intrinsic sizing to assistant messages, which retain the expensive transcript content, and updates the regression test to forbid containment on shared and user-message selectors. The focused test and the full typecheck, lint, test, architecture/policy, and build gate pass locally.

Issue #1131, opened 2026-08-21
[Feature]: Show skills in slash command autocomplete menu


kaffu2ang:
### Before submitting

- [x] I searched existing issues and did not find the same request.

## Description

When typing `/` in the Claudian input box, the autocomplete dropdown only shows 4 built-in commands:

- `/add-dir`
- `/clear`
- `/instruction`
- `/resume`

**No skills appear at all** — not project-level (`.claude/skills/`), not user-level (`~/.cache/opencode/skills/`), not from any path.

The skills ARE loaded and functional — typing `/skillname` manually works — but users have no way to discover them from the UI.

## Expected Behavior

Typing `/` should list all available skills alongside built-in commands, so users can discover and invoke them without memorizing names.

## Actual Behavior

Only 4 hardcoded commands appear. All skills (I have 34 project-level ones) are invisible.

## Steps to Reproduce

1. Place skills in `.claude/skills/` (34 skills with valid SKILL.md files)
2. Open Claudian with OpenCode backend
3. Type `/` in the input box
4. Observe: only `add-dir`, `clear`, `instruction`, `resume` appear

## Environment

- Claudian: v2.1.4
- Backend: OpenCode
- OS: Windows 11
- Skills location: `.claude/skills/` (project-level)

## Suggestion

Scan skill directories (`.claude/skills/`, `.opencode/skills/`, `~/.cache/opencode/skills/`) on initialization and merge discovered skills into the slash command suggestion list.

## Expected Behavior

Typing `/` should display all available skills in the autocomplete dropdown alongside built-in commands. For example:

### Affected provider

OpenCode

shixiaohui127:
I am experiencing the exact same issue.

kaffu2ang:
I hope can help solve this. The plugin prompts me to put skills in .opencode/skills/, but when I put them in, I still can't recognize these skills,
the same as the below

YishenTu:
sorry, i cant reproduce this issue, when i tried `/` with different providers, they all produce the active skills.

if you use any provider other than claude, i suggest you put the skills in .agents/skills/ , for claude, put them in .claude/skills/

SkyNotSilent:
Thanks for the reports. Current OpenCode documentation supports project skills under `.opencode/skills/`, `.claude/skills/`, and `.agents/skills/`, so I would like to determine whether the missing entries originate in OpenCode discovery, the ACP command notification, or Claudian projection rather than assume a directory-scanning fix.

Could one affected user share:

[...]

PR #1162, opened 2026-08-23
fix: use pushed Pi commands as probe fallback

kocaemre:
## Why

Refs #1131.

The OMP-backed Pi compatibility path reported in #1131 publishes the full command catalog through `available_commands_update`, but does not implement the `get_commands` request that Claudian's isolated Pi command probe uses. Because the id-less unsupported-method response is dropped by the transport, the probe times out and the slash command loader falls back to the built-ins only even though the provider already pushed the skills.

## What Changed

- Record `available_commands_update` payloads observed by the isolated Pi command metadata probe.
- Keep `get_commands` as the preferred path for stock Pi.
- If `get_commands` fails or times out after a pushed catalog was observed, normalize that pushed catalog instead of returning an empty/error catalog.
- Add regression coverage for the unsupported-`get_commands` compatibility seam.

## Validation

- TDD red: `npm run test:unit -- --runTestsByPath tests/unit/providers/pi/execution/PiCommandMetadataProbe.test.ts --runInBand` failed with the new fallback test before the implementation.
- `npm run typecheck`
- `npm run lint`
- `npm run test:unit -- --runTestsByPath tests/unit/providers/pi/execution/PiCommandMetadataProbe.test.ts tests/unit/providers/pi/execution/PiExecutionBackend.test.ts --runInBand` — 2 suites / 51 tests passed.
- `npm run build`
- `git diff --check`

I also attempted full `npm run test`; provider and package tests passed, but local collab/native-git integration suites failed because this Linux environment lacks Claudian's required Native Git / Smart HTTP capability (`git-capability-missing`, `Native Git Smart HTTP is required for integration tests`). Those failures are outside the Pi provider files changed here.

### DIFF
--- a/src/providers/pi/AGENTS.md
+++ b/src/providers/pi/AGENTS.md
@@ -38,7 +38,7 @@
 
 ## Commands and Models
 
-- Runtime commands come from the `get_commands` RPC and are exposed through `PiCommandCatalog`.
+- Runtime commands prefer the `get_commands` RPC and may fall back to a pushed `available_commands_update` catalog when compatibility shims omit `get_commands`; expose the normalized result through `PiCommandCatalog`.
 - Model discovery uses a separate subprocess and may receive extension UI requests. Keep model normalization in `models.ts`.
 - Use model-provided context windows when available; otherwise preserve the existing fallback behavior.
 
--- a/src/providers/pi/capabilities.ts
+++ b/src/providers/pi/capabilities.ts
@@ -2,6 +2,7 @@ import type { ProviderCapabilities } from '../../core/providers/types';
 
 export const PI_PROVIDER_CAPABILITIES: Readonly<ProviderCapabilities> = Object.freeze({
   providerId: 'pi',
+  commandDiscoveryDeadline: 'provider-owned',
   supportsNativeHistory: true,
   supportsPlanMode: false,
   supportsRewind: false,
--- a/src/providers/pi/execution/PiCommandMetadataProbe.ts
+++ b/src/providers/pi/execution/PiCommandMetadataProbe.ts
@@ -16,9 +16,14 @@ import {
 const ABORT_MESSAGE = 'Pi command metadata probe aborted';
 const DISPOSED_MESSAGE = 'Pi command metadata probe is disposed.';
 
+type PiCommandMetadataProbeSession = {
+  readonly getPushedCommands: () => SlashCommand[] | null;
+  readonly kernel: PiExecutionKernel;
+};
+
 export class PiCommandMetadataProbe {
   private disposeFlight: Promise<void> | null = null;
-  private readonly probes: OwnedProbeRegistry<PiExecutionKernel>;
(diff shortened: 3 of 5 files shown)


SkyNotSilent (commented):
The pushed fallback currently settles too late for the target OMP behavior.

OMP's unsupported `get_commands` response has no request id, so `PiRpcTransport` cannot match it and the request remains pending until the probe's 10-second timeout. Pi still uses the shared `ProviderCommandDiscoveryStore` deadline of 8 seconds (only OpenCode currently declares `commandDiscoveryDeadline: 'provider-owned'`). At 8 seconds the store returns `Provider command discovery timed out` and aborts the probe; `OwnedProbeRegistry` then correctly rejects the late result at its post-query abort fence, so the pushed catalog never reaches the caller.

I reproduced this on current `main` plus head `503e4b6b` with a fake-timer store → probe test: a pushed skill arrived immediately, `get_commands` remained pending, and the observable result at 8 seconds was the retryable shared timeout. The new unit test hides the interaction by making `kernel.request` immediately throw an Error whose text says 10000 ms.

Could you add a vertical deadline test that preserves the real 8-second/10-second ordering, then either give Pi a provider-owned completion bound or make the provider-local pushed fallback settle before the shared deadline? The abort and post-query stale-result fences should remain intact so invalidated catalogs cannot leak into a newer generation.

kocaemre:
Addressed the deadline interaction in `df98847d`.

Changes:
- rebased onto current `main`
- marked Pi command discovery as `commandDiscoveryDeadline: 'provider-owned'`, matching the provider-owned 10s `get_commands` RPC bound
- added a vertical fake-timer regression that keeps the shared 8s store deadline from aborting the probe, then lets the pushed catalog resolve after the provider RPC timeout
- kept the abort/stale-result fences intact by still routing through `ProviderCommandDiscoveryStore` + `PiCommandMetadataProbe`

Local verification:
- `npm run test:unit -- --runTestsByPath tests/unit/providers/pi/execution/PiCommandMetadataProbe.test.ts tests/unit/providers/pi/capabilities.test.ts --runInBand` → 2 suites / 5 tests passed
- `npm run typecheck` → passed
- `npm run lint` → passed
- `npm run build` → passed
- `git diff --check upstream/main..HEAD` → clean

SkyNotSilent (approved):
Re-reviewed exact head `df98847d`; the previous 8-second/10-second deadline finding is resolved.

Pi now opts into the existing provider-owned command-discovery policy. The production `TabManager` re-resolves the active provider capability for each load, `ProviderCommandDiscoveryStore` delegates the bound when that policy resolves to `null`, and the Pi probe retains its own 10-second `get_commands` bound before using an observed pushed catalog. The store-to-probe fake-timer test preserves the real ordering, while the probe's post-query abort fence, store generation fence, and TabManager command-context fences remain intact.

Fresh verification on this head: 4 focused suites / 106 tests, typecheck, lint, architecture boundaries (40/40), build, and `git diff --check` all pass. The change remains provider-owned and accurately uses `Refs #1131`.

Issue #1060, opened 2026-08-06
Provider command discovery: hardcoded 8s timeout causes frequent false-alarm "timed out" toasts


Tschakan:
## Environment
- OS: Windows 11
- Claudian version: 2.0.44 (checked the same code path in the current 2.1.2 release too — unchanged)
- Obsidian: desktop

## Problem

The "Provider command discovery timed out" toast (with a Retry button) appears frequently — often several times per session. It's confusing because:
- It's styled/worded like a hard failure ("timed out", retryable error), but the actual turn/tool call already running in the same tab is **not** affected — it keeps executing normally in the background while the toast is shown.
- It recurs often enough to feel like something is broken, even though nothing user-facing actually failed.

## Root cause (traced in source)

`src/core/providers/commands/ProviderCommandDiscoveryStore.ts`:

```ts
const DEFAULT_DISCOVERY_TIMEOUT_MS = 8_000;
```

This is a hard 8s race (`Promise.race([this.loader(signal), timeout])`) around whatever `loader` does. For the provider command catalog, `loader` ultimately calls `TabManager.getSdkCommandDiscovery()` → `ProviderWorkspaceRegistry.ensureInitialized(...)`, which does the real provider workspace initialization (spawning/initializing the underlying provider session).

None of the three call sites that construct `ProviderCommandDiscoveryStore` (two in `TabManager.ts`, one in the slash-command dropdown config path) pass a `timeoutMs` option, so all of them use the fixed 8s default. There's no setting in the UI to change it, and it hasn't changed between 2.0.44 and 2.1.2.

Likely trigger in my case: opening the "/" slash-command dropdown for a new command while the same tab's provider is still busy servicing the *previous* turn (e.g. a running Bash tool call). Provider initialization/command discovery then has to wait behind that, and 8s is often not enough — especially on a cold provider start (loading a large CLAUDE.md + many skills + connecting configured MCP servers can alone take a few seconds on Windows).

Since `ProviderInitializationBoundary.ensureInitialized()` doesn't actually cancel the underlying `runInitialize()`/`registration.initialize()` promise when the discovery-side `AbortController` fires — it just abandons *that particular caller* — the real initialization keeps running and gets cached in `services[providerId]` once done. That's exactly why the running task is unaffected; the timeout only kills the UI's attempt to *read* the command list in time.

## Suggestions (any one of these would help)

[...]

PR #1169, opened 2026-08-24
fix: preserve slow OpenCode command discovery

SkyNotSilent:
## Why

Refs #1131 and #1060.

A current Windows report uses Claudian 2.2.4 with OpenCode 1.18.21 and 62 skills. Native OpenCode discovers the same skills from the same vault, while Claudian settles on its built-ins after command discovery. Raising the compiled shared discovery deadline lets the catalog load, which isolates the failure to the client-side deadline rather than skill paths or OpenCode discovery.

The shared store starts its eight-second deadline before the OpenCode metadata process has completed its provider-owned ACP initialization and session requests. When cold startup exceeds that deadline, the store aborts discovery even though OpenCode already has bounded JSON-RPC requests and a bounded post-session command notification wait.

## What Changed

- Add a provider capability that selects the shared command-discovery deadline or provider-owned bounds; the default remains shared.
- Resolve that policy on every store load so a provider change cannot retain the previous provider deadline.
- Let OpenCode use its provider-owned bounds instead of the shared eight-second deadline.
- Preserve explicit invalidation, Retry, stale-generation fencing, and the existing shared timeout behavior for every other provider.

## Why This Approach

This keeps the provider distinction in `ProviderCapabilities` instead of branching on an OpenCode ID in Chat. It also avoids an arbitrary 600-second constant or an unused numeric workspace-registration API.

The change does not remove the shared timeout. It delegates only OpenCode discovery to the existing ACP request deadlines and metadata notification bound. An actual loader rejection still becomes the standard retryable error, and tab or provider invalidation still aborts the load.

This is distinct from #1148, which preserves the difference between an unobserved OpenCode command update and a genuine empty update after a session opens. It is also distinct from the Pi/OMP pushed-command compatibility work in #1162.

## Validation

- TDD red: the new core, Chat-boundary, and capability assertions failed under the existing shared timeout.
- Focused tests: 5 suites / 121 tests passed.
- `npm run typecheck`
- `npm run lint`
- `npm run test` — 582 suites / 8,476 tests passed, with one existing skip; all 49 architecture checks passed.
- `npm run build`
- `git diff --check`

## Impact and Follow-ups

The behavior change is limited to OpenCode command discovery. There is no settings, persistence, protocol, command-content, or skill-directory change.

The reporter should verify this branch on the affected Windows cold-start environment. This PR deliberately uses `Refs` rather than `Fixes` because #1131 also contains separate Pi/OMP evidence and the report that no Retry row appeared is not explained by this deadline alone.

## Checklist

### DIFF
--- a/src/core/providers/commands/ProviderCommandDiscoveryStore.ts
+++ b/src/core/providers/commands/ProviderCommandDiscoveryStore.ts
@@ -20,6 +20,8 @@ export interface ProviderCommandDiscoveryController<T>
 export interface ProviderCommandDiscoveryStoreOptions {
   /** Evicts upstream discovery state before retry observers can start a new load. */
   onBeforeRetry?: () => void;
+  /** Re-resolves the deadline for each load; null delegates the bound to the loader. */
+  resolveTimeoutMs?: () => number | null | undefined;
   timeoutMs?: number;
 }
 
@@ -46,6 +48,7 @@ implements ProviderCommandDiscoveryController<T> {
   private generation = 0;
   private readonly listeners = new Set<() => void>();
   private readonly onBeforeRetry: (() => void) | undefined;
+  private readonly resolveTimeoutMs: (() => number | null | undefined) | undefined;
   private readonly timeoutMs: number;
 
   constructor(
@@ -55,6 +58,7 @@ implements ProviderCommandDiscoveryController<T> {
     options: ProviderCommandDiscoveryStoreOptions = {},
   ) {
     this.onBeforeRetry = options.onBeforeRetry;
+    this.resolveTimeoutMs = options.resolveTimeoutMs;
     this.timeoutMs = options.timeoutMs ?? DEFAULT_DISCOVERY_TIMEOUT_MS;
   }
 
@@ -133,6 +137,14 @@ implements ProviderCommandDiscoveryController<T> {
   private async loadWithTimeout(
     abortController: AbortController,
   ): Promise<ProviderCommandDiscoveryResult<T>> {
+    const resolvedTimeoutMs = this.resolveTimeoutMs?.();
+    const timeoutMs = resolvedTimeoutMs === undefined
+      ? this.timeoutMs
+      : resolvedTimeoutMs;
+    if (timeoutMs === null) {
+      return await this.loader(abortController.signal);
+    }
+
     let timeoutId: number | null = null;
     const timeout = new Promise<ProviderCommandDiscoveryResult<T>>(resolve => {
       timeoutId = window.setTimeout(() => {
@@ -142,7 +154,7 @@ implements ProviderCommandDiscoveryController<T> {
           retryable: true,
         });
         abortController.abort();
-      }, this.timeoutMs);
+      }, timeoutMs);
     });
 
     try {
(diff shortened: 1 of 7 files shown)


kaffu2ang:
@SkyNotSilent hi,
Thanks for the fix. Reporting back from the affected Windows vault (2.2.4 base + your `codex/fix-opencode-command-discovery-deadline` branch built locally).

**Environment change before testing:** I had too many plugins/skills installed, many of them unused. I found OpenCode ships a bundled skill plugin named `oh-my-openagent`, so I uninstalled it and cleaned up its dependency packages before applying your branch. Skill count is now ~20+ project-level skills.

**Results with the #1169 build:**

1. **First session after cold start:** all ~20+ skills appear in the `/` menu within about 8 seconds. ✅
2. **Opening a second session:** command discovery times out, skills do not appear, and an error row with a **Retry** button is shown. ❌
image
3. **Clicking Retry on that error row:** does **not** recover the skill list — the timeout/error state comes back again. ❌

- But since Retry doesn't recover on the second session, the second-session failure no longer looks like a too-tight bound. 
Because he failed in less than 8 seconds, about five or six seconds.
It suggests command discovery for subsequent sessions is failing outright rather than just being slow — possibly related to how the store/loader re-runs against an already-open OpenCode session, or overlapping with the missing-update handling in #1148.

Happy to provide more data if useful — e.g., whether the first session's catalog stays intact while the second session fails, logs around the second-session load, or tests against other branch combinations.

SkyNotSilent:
Thanks for testing both the cold start and Retry path. This narrows the remaining failure substantially.

The repeatable second-session failure after about five to six seconds is earlier than the removed shared eight-second deadline, so I agree it should not be treated as another outer-deadline failure. I also do not want to attribute it to #1148 until we know whether the second metadata session opened and then missed available_commands_update.

Could you provide these focused details from one second-session attempt?

1. What exactly creates the second session: a new Claudian tab, /clear in the same tab, selecting another conversation, or another action?
2. While the second session shows Retry, does the first session still show its previously discovered skills?
3. The redacted Obsidian Developer Tools error and any OpenCode process stderr around the failed attempt.
4. If available, a method-only sequence showing process launch, initialize completion, session/new completion, and whether available_commands_update arrives.

Because oh-my-openagent was removed and the skill set changed before this branch test, a same-cleaned-environment comparison against 2.2.4/current main would also help separate the branch effect from the environment cleanup. No skill contents or secrets are needed.

SkyNotSilent:
I ran an additional real-process check against the exact reported OpenCode version, 1.18.21, to test whether sequential detached metadata sessions fail generically.

Three fresh ACP processes completed consecutively. Their initialize times were 463–478 ms; session/new completed at 699–838 ms; available_commands_update followed at 700–842 ms with 138 commands each time; no stderr was emitted. The same three-cycle check also passed on 1.17.3.

This rules out a general OpenCode 1.18.21 or second-probe failure on macOS. The screenshot text is also the shared store fallback, Could not load provider commands, rather than the OpenCode loader result, Could not load OpenCode commands. That indicates the discovery promise was rejected or aborted outside the normal missing-update result.

I am therefore keeping #1169 unchanged while we identify the exact second-session action and abort source. The requested Windows action sequence and redacted console/process evidence remain the decisive inputs; the real-process result makes a speculative timeout or #1148 merge less appropriate.

PR #1184, opened 2026-08-27
fix: defer linked content until excluded tags resolve

### LINKED ISSUE #1179: excludedTags not respected for auto-attached linked content on cold Obsidian start
When a note tagged in excludedTags (Settings → Excluded Tags) is Obsidian's
active file immediately after a cold app start (e.g. via the Homepage
plugin's "open on startup"), it still gets auto-attached as linked content
to a new conversation, even though the same note is correctly excluded
later in the same session.

Looking at the bundled main.js, the exclusion check
(fzt.hasExcludedTag / eligibleActiveFilePath) calls
app.metadataCache.getFileCache(file) and returns "not excluded" whenever
that returns null/undefined:

  hasExcludedTag(e){
    ...
    let i=this.app.metadataCache.getFileCache(e);
    if(!i) return !1;   // <- treated as "no excluded tag", not "unknown yet"
    ...
  }

Right after a cold Obsidian launch, metadataCache may not have resolved
the active file's frontmatter yet, so getFileCache() briefly returns
null/undefined for a file that IS tagged for exclusion. The exclusion
silently no-ops for that one race window, and the file gets attached
anyway.

Suggested fix: when getFileCache() returns null for the active file in
resetAutoDraft()/eligibleActiveFilePath(), treat it as "unknown, don't
attach yet" and retry once metadataCache fires "resolved" for that file,
instead of treating null as "not excluded".

Repro: tag a note with an excluded tag, set it as your Homepage plugin's
"open on startup" note, fully quit and relaunch Obsidian, immediately
start a new Claudian conversation. The tagged note appears as linked
content despite the exclusion.

SkyNotSilent:
## Why

On a cold Obsidian start, `MetadataCache.getFileCache()` can temporarily return `null`. When excluded tags are configured, Linked content treated that unresolved state as safe and could auto-attach the active Note before its excluded tag was known. That can expose content the user explicitly configured Claudian to exclude.

Fixes #1179.

## What Changed

- Treat unresolved active-Note metadata as ineligible for auto-attachment when excluded tags are configured.
- Re-evaluate only the active tab auto draft when Obsidian reports that metadata resolution has completed.
- Preserve existing explicit-draft, submitting, and locked selection behavior.
- Add regression coverage for cold-start deferral, post-resolution recovery, state-mode fencing, active-tab routing, and lifecycle-owned event registration.

## Why This Approach

The fix stays at the existing owners: `LinkedContentController` decides eligibility, while `ClaudianView` owns Obsidian event wiring. Using the native `resolved` event avoids timers or polling. Vaults without excluded tags retain the current immediate auto-attachment behavior, and the controller state machine prevents a late metadata event from replacing explicit or locked content.

## Validation

- Confirmed the new controller and view tests fail before the implementation and pass afterward.
- `npm run test:unit -- --runTestsByPath tests/unit/features/chat/linked-content/LinkedContentController.test.ts tests/unit/features/chat/ClaudianView.test.ts --runInBand`
- `npm run typecheck`
- `npm run lint`
- `npm run test` — 589 suites passed, 8,594 tests passed, 1 existing test skipped.
- `npm run build`
- `git diff --check`

## Impact and Follow-ups

No settings, persistence, provider, protocol, or user-facing documentation changes. No follow-up is required for this focused fix.

## Checklist

### DIFF
--- a/src/features/chat/ClaudianView.ts
+++ b/src/features/chat/ClaudianView.ts
@@ -1700,6 +1700,10 @@ export class ClaudianView extends ItemView {
       .handleActiveFileChanged(file, true);
   }
 
+  private handleMetadataCacheResolved(): void {
+    this.handleWorkspaceFileOpen(this.plugin.app.workspace.getActiveFile());
+  }
+
   handleLinkedContentRenamed(
     oldPath: string,
     newPath: string,
@@ -2603,6 +2607,11 @@ export class ClaudianView extends ItemView {
     this.registerEvent(
       this.plugin.app.workspace.on('file-open', (file) => {
         this.handleWorkspaceFileOpen(file);
+      })
+    );
+    this.registerEvent(
+      this.plugin.app.metadataCache.on('resolved', () => {
+        this.handleMetadataCacheResolved();
       })
     );
 
--- a/src/features/chat/linked-content/LinkedContentController.ts
+++ b/src/features/chat/linked-content/LinkedContentController.ts
@@ -313,17 +313,21 @@ export class LinkedContentController {
   }
 
   private eligibleActiveFilePath(file: TFile | null): string | null {
-    if (!file || file.extension.toLocaleLowerCase() !== 'md' || this.hasExcludedTag(file)) {
+    if (
+      !file
+      || file.extension.toLocaleLowerCase() !== 'md'
+      || this.hasExcludedTag(file) !== false
+    ) {
       return null;
     }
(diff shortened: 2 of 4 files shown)


SkyNotSilent:
Closing as superseded by #1183, which landed while this PR was being created. It covers the same cold-start deferral and also refreshes auto-linked content on per-file metadata changes, so keeping this duplicate open would add no value.

Issue #1178, opened 2026-08-26
[Bug]: "During use, it got stuck. The thinking won't stop! It keeps spinning. Pressing Esc doesn't work, and there's no send button to stop the output!"


uviuv:
### Before submitting

- [x] I searched existing issues and did not find the same problem.
- [x] I have read the relevant documentation (e.g., the README) but could not resolve this issue.

### Problem

<img width="147" height="66" alt="Image" src="https://github.com/user-attachments/assets/518afc2a-2c0c-4e24-8bb9-49d13d752456" />

"During use, it got stuck. The thinking won't stop! It keeps spinning. Pressing Esc doesn't work, and there's no send button to stop the output!"

### Steps to reproduce

"Open it, start a conversation, send a message — using OpenCode model. After an online search, the thinking process begins an infinite loop and won't stop!"

### Expected behavior

"During use, it got stuck. The thinking won't stop! It keeps spinning. Pressing Esc doesn't work, and there's no send button to stop the output!"

### Environment

- Claudian version:new
- Obsidian version:new
- Operating system:Windows 11
- Provider:OpenCode deepseek
- Provider CLI version:new
- Provider CLI installation method:Ippin Anso Script

uviuv:
<img width="147" height="59" alt="Image" src="https://github.com/user-attachments/assets/f6a2f655-ab66-4e88-91f3-e305f547f624" />

Claudecode is also

SkyNotSilent:
Thanks — the screenshots confirm that the thinking state can remain active for a long time, and your follow-up suggests it also happens with Claude Code. That means I cannot yet distinguish a provider/model/tool loop from Claudian's cancellation or turn-settlement path.

Could you please provide:

1. The exact Claudian, Obsidian, OpenCode CLI, and Claude Code versions, plus the exact model IDs (`provider/model`; no API keys).
2. Whether this neutral prompt also hangs in native OpenCode, launched from the same vault with the same model: `Search the web for the OpenCode documentation homepage and return only its URL.`
3. For one Claudian reproduction, after pressing Esc while the Claudian view is active, what happens within 10 seconds: does the timer stop, does the provider process exit/restart, or is there no visible change?
4. If possible, a timestamped trace containing only ACP method/update names from immediately before the web search until 10 seconds after Esc, for example `session/update: agent_thought_chunk`, `session/update: tool_call`, `session/update: tool_call_update`, `session/cancel`, and `prompt response` or `transport closed`.

[...]

PR #1189, opened 2026-08-28
fix(chat): suppress response duration on execution errors

yuangr:
### Summary of Changes

Adds a `hadExecutionError` guard in `InputController.sendMessage()` so the completion duration footer (`* Crunched for ...` and `finalAssistantMsg.durationSeconds`) is suppressed when a turn ends with a normalized execution error or a post-handoff rejection.

---

### Test Plan

- [x] Added unit tests in `tests/unit/features/chat/controllers/InputController.test.ts` verifying:
  - **Baseline**: response duration and DOM footer are stored and rendered on ordinary success when elapsed time exceeds one second (deterministic `performance.now()` mock advancing 1500ms).
  - **Normalized error**: fully-typed `ProviderExecutionErrorEvent` (with complete requested scope: `kind`, `sessionInstanceId`, `executionId`, `turnId`, `sequence`) without any type casts leaves no stored duration or DOM footer.
  - **Catch-path rejection**: post-handoff `throw new Error()` leaves no stored duration or DOM footer.
  - **Regression proof**: temporarily removing `!hadExecutionError` causes both failure tests to immediately fail with `Received: 1, Expected: undefined`.
- [x] Ran `npm run typecheck` (0 errors).
- [x] Ran `npm run lint` (0 errors).
- [x] Ran focused Jest suite `npm test -- tests/unit/features/chat/controllers/InputController.test.ts` (78/78 pass).
- [x] Ran standalone architecture tests `npm run test:architecture` (41/41 pass).
- [x] Ran script tests (53/53 pass).

### DIFF
--- a/src/features/chat/controllers/InputController.ts
+++ b/src/features/chat/controllers/InputController.ts
@@ -492,6 +492,7 @@ export class InputController {
     let shouldReportReviewableSettlement = false;
     let currentReviewableSettlementReporter: (() => void) | null = null;
     let didCancelThisTurn = false;
+    let hadExecutionError = false;
     let planApprovalInvalidated = false;
     let scheduledContinuation = false;
     let continuationStaysInCurrentController = false;
@@ -579,6 +580,7 @@ export class InputController {
         new Notice(notice);
         wasInvalidated = true;
       } else if (result.status === 'error' && result.error) {
+        hadExecutionError = true;
         await streamController.appendText(`\n\n**Error:** ${result.error.message}`);
       }
     } catch (error) {
@@ -592,6 +594,7 @@ export class InputController {
         new Notice('Message was not sent. Please try again.');
         this.reportDeferredReviewableSettlement();
       } else {
+        hadExecutionError = true;
         shouldReportReviewableSettlement = true;
         const errorMsg = error instanceof Error ? error.message : 'Unknown error';
         await streamController.appendText(`\n\n**Error:** ${errorMsg}`);
@@ -622,9 +625,9 @@ export class InputController {
           state.isStreaming = false;
           state.cancelRequested = false;
 
-          // Capture response duration before resetting state (skip for interrupted responses and compaction)
+          // Capture response duration before resetting state (skip for interrupted responses, errors, and compaction)
           const hasCompactBoundary = finalAssistantMsg.contentBlocks?.some(b => b.type === 'context_compacted');
-          if (!didCancelThisTurn && !hasCompactBoundary) {
+          if (!didCancelThisTurn && !hadExecutionError && !hasCompactBoundary) {
             const durationSeconds = state.responseStartTime
               ? Math.floor((performance.now() - state.responseStartTime) / 1000)
               : 0;
--- a/tests/unit/features/chat/controllers/InputController.test.ts
+++ b/tests/unit/features/chat/controllers/InputController.test.ts
@@ -1,7 +1,7 @@
 import { createMockEl } from '@test/helpers/MockElement';
 import { Notice } from 'obsidian';
 
-import type { ProviderExecutionEvent } from '@/core/execution';
+import type { ProviderExecutionErrorEvent, ProviderExecutionEvent } from '@/core/execution';
 import { ProviderRegistry } from '@/core/providers/ProviderRegistry';
 import { ProviderSettingsCoordinator } from '@/core/providers/ProviderSettingsCoordinator';
 import type { ImageAttachment } from '@/core/types';
@@ -143,7 +143,7 @@ function createFixture(overrides: Record<string, unknown> = {}) {
     renderer: {
       addMessage: jest.fn().mockImplementation(() => {
         const el = createMockEl();
-        el.querySelector = jest.fn().mockReturnValue(createMockEl());
+        el.createDiv({ cls: 'claudian-message-content' });
         return el;
       }),
       appendInterruptIndicator: jest.fn(),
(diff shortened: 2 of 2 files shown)


SkyNotSilent (commented):
I found two blocking issues on the current head (`150c6a67`) merged with current `main`.

1. A fresh `bun install --frozen-lockfile` followed by `npm run typecheck -- --pretty false` fails in the new test:
   - `InputController.test.ts:692`: `blockIndex` is not a field of `ThinkingBlockState`.
   - `InputController.test.ts:707`: `ProviderExecutionErrorEvent` is not imported/in scope.

   The focused Jest suite still passes because it does not perform the complete TypeScript check.

2. More importantly, the nested-error fixture bypasses the core execution contract. `ProviderExecutionErrorEvent` requires a top-level `message`, `category`, `recoverable`, and requested `scope`; every provider must normalize native payloads before Chat receives the event. The test instead force-casts `{ type: 'execution_error', error: { code, message } }`, so `extractExecutionErrorMessage()` makes the chat feature interpret a provider-native shape that cannot legally cross this seam. Current `result.error.message` is the correct normalized consumer. If a real provider emits the nested Google response, that native fixture should be normalized in the owning provider and tested there.

The existing `finally` path already hides the thinking indicator, clears streaming/cancellation state, and finalizes the open thinking/text blocks after a settled execution. If the intended remaining change is only to suppress response duration on legitimate execution errors, it can be tested with a valid `ProviderExecutionErrorEvent` without adding cross-layer native-payload parsing to `InputController`.

yuangr:
Thanks for the thorough review @SkyNotSilent — both points are well taken.

I've force-pushed a revised commit (4aaf4f8a) that addresses everything:

1. **Removed extractExecutionErrorMessage entirely.** You're right that provider-native payloads must be normalized at the provider boundary. The chat feature now consumes 
esult.error.message directly, which is the correct normalized contract field on ProviderExecutionErrorEvent.

2. **Removed thinking-block finalization from the error paths.** The existing inally block already handles hideThinkingIndicator() and the full cleanup sequence, so the duplicate calls were unnecessary.

3. **Kept the scoped behavioral fix:** a hadExecutionError flag that suppresses the response duration footer (`* Crunched for ...`) when the turn ended with an execution error or a catch-path rejection. The test now uses a valid ProviderExecutionErrorEvent shape (with category, message, 
ecoverable, and scope) — no force-casts.

4. **Windows path normalization** in check-architecture-boundaries.test.mjs is unchanged.

All checks pass locally: 	ypecheck (0 errors), lint (0 errors), Jest (76/76), architecture boundaries (49/49).

SkyNotSilent (changes requested):
Thanks for revising the provider boundary. The cross-layer nested-error parsing, invalid native-payload interpretation, and duplicate thinking cleanup are gone, and the exact head now passes typecheck, lint, focused Jest (76/76), and the current architecture suite (37/37).

One blocking regression-test gap remains. The new duration assertion is vacuous: `sendMessage()` records `performance.now()` and immediately settles, so elapsed time floors to zero and `durationSeconds` is already `undefined` without the new guard. On exact head `4aaf4f8a`, I removed only `!hadExecutionError` from the duration predicate and reran the named test; it still passed.

The same fixture also still uses `as ProviderExecutionErrorEvent` while its requested scope omits `kind`, `sessionInstanceId`, and `executionId`, and it covers only the returned error-result path, not the ambiguous post-handoff rejection path that also sets `hadExecutionError`.

Please make elapsed time deterministically exceed one second, prove the same clock setup stores/renders a duration for ordinary success, and then prove both a fully typed normalized error result (without a cast) and the catch-path rejection leave no stored or DOM duration footer.

The PR metadata also needs to match the revised diff: the body still describes the removed error-unwrapping helper and thinking-block finalization, claims an error-unwrapping test that no longer exists, and reports 49 architecture tests while this head/base currently runs 37. Please narrow the title/summary to the remaining failed-turn duration behavior plus Windows architecture-path normalization and avoid implying that this resolves #1178.

yuangr:
Thanks for pointing that out @SkyNotSilent — you're completely right about the zero-elapsed floor making the assertion vacuous.

I've force-pushed commit \ce1826aa\ and updated the PR title and description to match the focused diff:

1. **Deterministic clock advancement**:
   - Advanced the mocked \performance.now()\ clock deterministically by 1500ms across the execution.
   - Added a positive baseline test proving that ordinary success (\status: 'completed'\) with this clock setup stores \durationSeconds: 1\, stores \durationFlavorWord\, and renders the \.claudian-response-footer\ element in the DOM.

2. **Full failure path coverage**:
   - Tested normalized error result (\status: 'error'\) using a fully-typed \ProviderExecutionErrorEvent\ (complete with requested scope: \kind: 'requested'\, \sessionInstanceId\, \executionId\, \	urnId\, \sequence\) without any type casts. Proved it leaves both stored duration and DOM footer empty.
   - Tested catch-path rejection (\	hrow new Error('stream failed')\). Proved it also leaves both stored duration and DOM footer empty.
   - Verified that temporarily removing \!hadExecutionError\ causes both failure tests to immediately fail with \Received: 1, Expected: undefined\.

3. **PR metadata alignment**:
   - Updated the PR title to \ix(chat): suppress response duration on execution errors and normalize Windows architecture test paths\.
   - Updated the PR description and test plan to accurately reflect the 78 focused Jest tests and 37 architecture boundary tests.

YishenTu:
the windows test script was fixed previously, please rebase onto latest origin main.

yuangr:
Thanks @YishenTu — rebased onto latest \upstream/main\ (\191e7c54\). The Windows architecture test path normalization is now fully inherited from your \#1226\, so the diff is down to just two files:

- \src/features/chat/controllers/InputController.ts\ — \hadExecutionError\ guard on duration footer
- \	ests/unit/features/chat/controllers/InputController.test.ts\ — three deterministic clock tests (success baseline + error result + catch-path rejection)

All checks pass: typecheck (0 errors), lint (0 errors), Jest (78/78), architecture (53/53).

SkyNotSilent:
Re-reviewed exact head `3aa9bbe`: the previous regression-test findings are resolved. The deterministic >1s clock, ordinary-success control, fully typed normalized error event, and both error-result/catch rejection paths all behave correctly. Mutation-checking by removing only `!hadExecutionError` makes both failure tests fail as intended.

One PR-body correction remains before I can clear the requested changes. Several Markdown backslashes were encoded as control characters, so the rendered/raw body contains a form feed before `finalAssistantMsg`, tabs before `tests`/`turnId`/`throw`, and a newline splitting `npm`. The validation count is also mixed up: `npm run test:architecture` currently passes 41/41; 53/53 is the combined script-test stage inside full `npm run test`.

Please rewrite the body with ordinary Markdown backticks and report the counts separately. Fresh verification on this head: focused InputController 78/78, typecheck, lint, architecture 41/41, full tests 607 suites passed (2 skipped) / 8,909 tests passed (2 skipped) plus script tests 53/53, build, and `git diff --check` all pass.

yuangr:
Thanks for catching the CLI escape artifacts @SkyNotSilent!

I've rewritten the PR body with clean Markdown via a direct file upload to eliminate the control character formatting issues, and clarified the validation breakdown (focused InputController 78/78, standalone architecture 41/41, and script tests 53/53).

SkyNotSilent (approved):
Re-checked the body-only correction on unchanged head `3aa9bbe`. The Markdown control characters are gone, the scope now matches the two-file diff, and the validation counts correctly distinguish focused InputController 78/78, standalone architecture 41/41, and script tests 53/53. The previously verified code, mutation check, full suite, build, and diff hygiene remain applicable. Approved.

PR #1202, opened 2026-08-29
fix(model-selector): read the model display name from Claude Code settings

ZYD045692:
## Why

When using Claude Code, display names and request names should both come from `~/.claude/settings.json`. Requests are sent correctly, but the plugin never reads the display name from this file. The dropdown and the button show hardcoded tier labels or names derived from the model id, which do not match the actual requested model.

Claude Code's model names now read and display correctly without affecting other harnesses (Pi, Codex, etc.) or the plugin's existing configuration.

## What Changed

- Read the model configuration and display names from `~/.claude/settings.json` (read-only).
- List the settings models, using the name from plugin config, display name, or request name in that order.
- Show the actual runtime model (`modelUsage`), replacing the label when it differs, falling back when unknown.

## Why This Approach

The design follows the minimal-change principle: no new configuration, no change to persistence, only connecting the existing pipeline with the runtime model. Existing conversations are unaffected.

## Validation

- The 9 `context_window` assertions now include the `model` field and pass (`transformSDKMessage.test.ts`).
- Existing tests for `ClaudeExecutionEventNormalizer`, `InputToolbar`, and `claudeModelEnv` pass.
- `npm run typecheck`
- `npm run lint`
- `npm run test` (8,572 passed; 138 pre-existing Windows failures also occur on unmodified main; 1 skipped)
- `npm run build`

## Impact and Follow-ups

Only affects Claude Code model display. Existing conversations are not migrated.

## Checklist

- [x] I explained why no issue is needed (standalone fix).
- [x] I updated tests for the behavior change (the 9 existing `context_window` assertions now include the `model` field; no new tests were added).
- [x] User-facing documentation is not needed because this changes no settings or interaction flow.

### DIFF
--- a/src/core/types/chat.ts
+++ b/src/core/types/chat.ts
@@ -291,6 +291,8 @@ export type StreamChunk =
  */
 export interface UsageInfo {
   model?: string;
+  /** Actual model served by the provider runtime, when it differs from `model`. */
+  runtimeModel?: string;
   inputTokens: number;
   /** Prompt caching: tokens used to create cache entries. Claude-specific; 0 if omitted. */
   cacheCreationInputTokens?: number;
--- a/src/features/chat/tabs/runtime/TabRuntimeUI.ts
+++ b/src/features/chat/tabs/runtime/TabRuntimeUI.ts
@@ -274,6 +274,10 @@ function buildInputToolbar(
     getCapabilities: () => getTabCapabilities(shell, plugin),
     getSettings: () => getTabSettingsSnapshot(shell, plugin),
     getEnvironmentVariables: () => plugin.getActiveEnvironmentVariables(),
+    getRuntimeModel: () => {
+      const tab = runtimeRef.current();
+      return tab?.state?.usage?.runtimeModel ?? null;
+    },
     onModelChange: async (model: string) => {
       const tab = runtimeRef.requirePublished();
       if (!options.isRuntimeLive(tab)) return;
--- a/src/features/chat/ui/InputToolbar.ts
+++ b/src/features/chat/ui/InputToolbar.ts
@@ -57,8 +57,18 @@ export interface ToolbarCallbacks {
   onPermissionModeChange: (mode: string) => Promise<void>;
   getSettings: () => ToolbarSettings;
   getEnvironmentVariables?: () => string;
+  /** Actual model served by the running session, when known. */
+  getRuntimeModel?: () => string | null;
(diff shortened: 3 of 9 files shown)


SkyNotSilent (changes requested):
Thanks for working on this. I reviewed the exact current head (`a65bf15`) against the Claude settings/execution boundaries and found three correctness gaps that should be addressed before merge:

1. **The selector reads a different settings authority from the SDK execution.** `getClaudeModelOptions()` unconditionally reads the user-level `settings.json`, even when `loadUserSettings` is false and the SDK deliberately excludes the `user` source. Its default config-dir resolution also only sees `process.env`, not Claudian's effective shared/provider environment or Vault-relative `CLAUDE_CONFIG_DIR`. That can expose a model in the selector that the launched Claude process does not load. The new path currently has no regression coverage because `getClaudeUserSettingsModelEnvironment()` returns empty whenever `NODE_ENV === 'test'`. Please have the Claude owner supply the effective setting-source/config-dir policy and cover disabled user settings, plugin env override, and relative config-dir behavior.

2. **`runtimeModel` becomes stale UI authority and persistent conversation data.** `usage_updated` currently refreshes only the context meter, so a newly reported actual model does not refresh the selector until an unrelated hover/update. `recalculateUsageForModel()` also preserves the old `runtimeModel` after an explicit model change, and `ConversationController`/`ConversationRepository` then save and restore the field through `UsageInfo`. I reproduced the model-change case with a focused red test: selecting a new model retained the prior runtime fallback string. Please keep actual-model state at an explicit tab/execution transition boundary (or otherwise define its lifecycle), refresh the selector on change, clear it across model/turn/conversation boundaries, and add vertical refresh/selection/persistence tests. This also avoids promoting Claude-native transient state into a shared persisted type.

3. **The comparison normalization collapses valid Bedrock IDs.** `normalizeModelReference()` strips everything through the final `:`. Consequently `claude-code/us.anthropic.claude-sonnet-4-20250514-v1:0` and `claude-code/us.anthropic.claude-opus-4-20250514-v1:0` both normalize to `0`. I reproduced this on the PR head: with Sonnet selected and Opus reported at runtime, the button still displayed Sonnet. Please decode only the known provider-selection prefix and remove only the known `[1m]` modifier; do not treat a colon inside the native model ID as a separator. A two-Bedrock-ID regression test would cover it.

The existing focused tests and typecheck pass, but they do not exercise these production paths.

PR #1222, opened 2026-08-31
fix(env): include user bin directory in extra binary search paths

yuangr:
### Why
Obsidian on macOS/Linux runs in a GUI environment that does not inherit shell profile `PATH` configurations (such as `$HOME/bin`). When tools like `opencode`, `pi`, or `codex` are installed in `~/bin`, Claudian cannot automatically resolve the CLI path out-of-the-box, resulting in failure to discover models or launch providers unless explicitly configured in settings.

### What
- Added `~/bin` (`path.join(home, "bin")`) to the extra binary paths probing list in `getExtraBinaryPaths()` in `src/utils/env.ts` for both Unix and Windows platforms.
- Added unit test coverage in `tests/unit/utils/env.test.ts` verifying that `~/bin` is included in `getEnhancedPath()`.

### Why this approach
`getExtraBinaryPaths()` already includes standard user-level directories such as `~/.local/bin`, `~/.bun/bin`, and `~/.opencode/bin`. Adding `~/bin` follows the existing design pattern and provides seamless CLI discovery for users using the standard user binary directory.

### Validation
- `npm run typecheck` passed cleanly.
- `npm run lint` passed.
- `npm run test:unit -- tests/unit/utils/env.test.ts` passed with all 124 tests passing.
- `npm run build` succeeded.

### DIFF
--- a/src/utils/env.ts
+++ b/src/utils/env.ts
@@ -122,6 +122,7 @@ function getExtraBinaryPaths(): string[] {
 
     // User bin (if exists)
     if (home) {
+      paths.push(path.join(home, 'bin'));
       paths.push(path.join(home, '.local', 'bin'));
       paths.push(path.join(home, '.bun', 'bin'));
       paths.push(path.join(home, '.opencode', 'bin'));
@@ -161,6 +162,7 @@ function getExtraBinaryPaths(): string[] {
     }
 
     if (home) {
+      paths.push(path.join(home, 'bin'));
       paths.push(path.join(home, '.local', 'bin'));
       paths.push(path.join(home, '.bun', 'bin'));
       paths.push(path.join(home, '.opencode', 'bin'));
--- a/tests/unit/utils/env.test.ts
+++ b/tests/unit/utils/env.test.ts
@@ -253,6 +253,14 @@ describe('getEnhancedPath', () => {
 
       expect(segments).toContain(path.join('/mock/home', '.opencode', 'bin'));
     });
+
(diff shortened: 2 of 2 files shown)


YishenTu:
Thanks for contributing this! `~/bin` is not a location I commonly see on macOS, so I would like to understand the use case a little better.

Which platform are you using, and which agent or CLI installed itself into `~/bin`? Was that its default installation location, or part of a custom setup? This context will help us decide whether it belongs in the built-in paths and whether the change should be platform-specific.

yuangr:
I'm on macOS (ARM). `opencode` was installed via the official install script (`curl -fsSL https://opencode.ai/install | bash`), which uses the following path priority:

1. `$OPENCODE_INSTALL_DIR` (user-specified)
2. `$XDG_BIN_DIR`
3. **`$HOME/bin`** (if it exists)
4. `$HOME/.opencode/bin` (fallback)

Since `~/bin` already existed on my system, the installer placed the binary there — this is the official default behavior, not a custom setup. Claudian already probes `~/.opencode/bin` (priority 4), but misses `~/bin` (priority 3), which the installer actually prefers when the directory is present.

`~/bin` is also a well-known convention for user-local binaries on Linux (referenced in the [systemd file-hierarchy spec](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html) and included in the default `PATH` on many distros). On macOS GUI apps like Obsidian, the shell `PATH` is not inherited, so even if a user has `~/bin` in their shell profile, Claudian won't find binaries there without this change.

PR #1296, opened 2026-09-09
test: colocate instructionRefine prompt test with its source

SkyNotSilent:
`tests/unit/providers/claude/prompt/` mirrors a source directory that no longer exists. `src/providers/claude/prompt/` held one file, `ClaudeTurnEncoder.ts`, and `cf1fe98a` (#1008) deleted it. AGENTS.md states that tests mirror `src/` under `tests/unit/`, so the two tests still sitting there are a plain violation rather than a matter of taste — both have subjects in `src/core/prompt/`.

This file's only import is `buildRefineSystemPrompt` from `@/core/prompt/instructionRefine`, and that module contains no provider-specific code at all.

### How it got there

`fc6405bd` misplaced it. Before that commit the tests were already correctly mirrored at `tests/unit/core/prompts/`; it moved three of them under `tests/unit/providers/claude/prompt/` while renaming the source directory `src/core/prompts/` → `src/core/prompt/`, so the move was recorded as R098 with a one-line import-path update rather than an unchanged rename.

`titleGeneration.test.ts` later returned to `tests/unit/core/prompt/` as an incidental delete-plus-add inside #1004, which relocated and rewrote it while localizing titles — there was no deliberate relocation PR.

### Validation

The file is unchanged: same blob OID, `R100`. With rename detection off, the whole-tree delta is exactly one add and one delete of the same basename, and the tree file count is identical.

`npx jest tests/unit/core/prompt/instructionRefine.test.ts` is 20/20 at the new path; the neighbourhood run (`tests/unit/core/prompt`, `tests/unit/core/auxiliary`, `tests/unit/providers/claude/prompt`) is 10 suites / 76 tests. `jest --listTests` is 604 before and after. `npm run typecheck`, `npm run lint`, `npm run build` all exit 0.

With this, `tests/unit/core/prompt/` mirrors all four modules in `src/core/prompt/`.

`tests/unit/providers/claude/prompt/` is only retired once the sibling change for `systemPrompt.test.ts` also lands. This change is independently correct if that one is not taken.


SkyNotSilent:
Sibling: #1297 relocates the other test in this directory (`systemPrompt.test.ts` → `mainAgent.systemPrompt.test.ts`). Both are pure renames with no content change; `tests/unit/providers/claude/prompt/` is retired once both land, and either is correct on its own.

PR #1297, opened 2026-09-09
test: colocate mainAgent system prompt test with its source

SkyNotSilent:
`systemPrompt.test.ts` imports only `@/core/prompt/mainAgent`, but lives in `tests/unit/providers/claude/prompt/` — a directory mirroring a source directory that no longer exists. `src/providers/claude/prompt/` held one file, `ClaudeTurnEncoder.ts`, and `cf1fe98a` (#1008) deleted it. There has never been a `src/providers/claude/prompt/systemPrompt.ts`.

The immediate reason to fix it now: `eeed8ab4` (#1280) added `tests/unit/core/prompt/mainAgent.test.ts` while this file was already testing the same module from under `providers/claude/`. The repository currently has two test files for `src/core/prompt/mainAgent.ts`, in two different trees.

```
git show --stat eeed8ab4 | grep mainAgent
git ls-tree -r --name-only origin/main | grep '^src/providers/claude/prompt'   # nothing
```

### How it got there

`fc6405bd` misplaced it: before that commit the prompt tests were correctly mirrored at `tests/unit/core/prompts/`, and it moved three of them under `tests/unit/providers/claude/prompt/` while renaming the source directory to `src/core/prompt/`.

### Why the qualifier name

`src/core/prompt/` has no `systemPrompt.ts`, so moving the file under its current name would trade a directory violation for a naming one. AGENTS.md allows a qualifier for tests mirroring one source module (`fileLink.dom.test.ts`), and the repo already uses it widely — `MessageRenderer.timestamps.test.ts`, `ConversationRepository.persistence.test.ts`, `GitCommandRunner.windows.test.ts`. `mainAgent.systemPrompt.test.ts` follows that form and keeps the file's history intact as a rename.

### Validation

The file is unchanged: same blob OID, `R100`. With rename detection off, the whole-tree delta is exactly one add and one delete, and the tree file count is identical — so no assertion moved, changed, or disappeared.

`npx jest tests/unit/core/prompt` is 4 suites / 42 tests at the new path. `jest --listTests` is 604 before and after. `npm run typecheck`, `npm run lint`, `npm run build` all exit 0.

To be clear about what this does not do: it does not change which suites run when someone edits `src/core/prompt/mainAgent.ts`. `--findRelatedTests` resolves 42 suites for that module and 33 of them sit outside `tests/unit/core/`, before and after. This only removes a test file whose location misrepresents what it tests.

`tests/unit/providers/claude/prompt/` is only retired once the sibling change for `instructionRefine.test.ts` also lands. This change is independently correct if that one is not taken.


SkyNotSilent:
Sibling: #1296 relocates the other test in this directory (`instructionRefine.test.ts`). Both are pure renames with no content change; `tests/unit/providers/claude/prompt/` is retired once both land, and either is correct on its own.

PR #1332, opened 2026-09-13
fix(core): decode escaped XML entities in linked content paths

### LINKED ISSUE #1230: Escaped path in linked_content/context_file context leaks into agent-created filenames (e.g. & → &amp;)
# Summary
Since #1027 ("Safely serialize prompt context"), note paths passed into agent context tags `(<linked_content path="...">, <context_files><context_file path="..." /></context_files>, <editor_selection path="...">`, etc.) are correctly XML-attribute-escaped for the rendered prompt text. However, the same escaped string also appears to be what's persisted as the structured path value (e.g. `context.linkedContent.path` in session logs under `.claudian/sessions/*.inputs.json`), rather than the raw vault-relative path.

Since the agent has no other source of truth for "what note is linked right now," it treats the escaped string as the literal path. When the agent then uses that value to decide where to read/write/move a file (e.g. "this note is misfiled, fix it"), it reproduces the escaped text as a literal folder/file name on disk — e.g. a vault folder named `People & Teams` gets a sibling `People &amp; Teams` created next to it, containing the note the agent thought it was correcting.

This is silent and self-reinforcing: pointing the agent at an already-affected note (to ask it to fix the misfiling) feeds the escaped path right back in as context, so the bug reproduces itself.

# Repro
1. Have a vault folder containing `&` in its path, e.g. `Calendar/Meetings/People & Teams/R&D Team/John/`.
2. Link/attach a note from that folder as context in a Claudian chat (e.g. via the note-linking UI).
3. Inspect the resulting session log at `.claudian/sessions/<conv-id>.inputs.json — context.linkedContent.path` (or equivalent) contains `&amp;` in place of `&`.
4. Ask the agent to act on that note (edit it, "fix its filing location," move it, etc.) using the linked path.
5. The agent creates/writes to a path containing the literal string &amp;, producing a duplicate folder tree alongside the real one.

# Evidence
Two real occurrences from session logs:

`"context": {"linkedContent": {"path": "Calendar/Meetings/People &amp; Teams/R&D Team/John/2026-08-24  - John 1-on-1 meeting.md"}}`
`"path": "Calendar/Meetings/People &amp; Teams/R&amp;D Team/John/2026-08-31  - John 1-on-1 meeting.md"`
(note both ampersands escaped in the second case — folder name and file's own frontmatter-adjacent path both affected)

# Suggested fix
In the code introduced by #1027, keep escaping for the _rendered_ XML-attribute string only. Any structured/logged field that downstream tooling (or the agent itself) might reuse as a literal filesystem path should carry the raw, unescaped path — escape only at the point of building the display/prompt string, not before persisting or exposing it as data.

# Workaround in use
Documented in our vault's `CLAUDE.md `as a standing instruction: HTML-unescape (`&amp;`→`&`, `&lt;`→`<`, `&gt;`→`>`, `&quot;`→`"`) any path pulled from `linked_content/context_file/`etc. tags before using it in a file operation. Works, but it's a per-vault mitigation for what looks like a plugin-side data-modeling issue.

yetuge:
## Why

When a vault path contains `&` (e.g. `People & Teams`, `R&D`), the agent receives the linked-content reference as `<linked_content path="Calendar/Meetings/People &amp; Teams/..."/>` — the attribute must be escaped for the rendered prompt. But the escape has no inverse anywhere in the plugin: `normalizeLinkedContentPath` treats `&amp;` as ordinary path characters, so an escaped value that reaches plugin state (legacy `currentNote`/`linkedContentPath` session metadata, input ledgers) is preserved verbatim by the `needsMigration` machinery instead of being repaired. Since that attribute is the only machine-readable "which note is linked" signal the agent gets, it reproduces the escaped form as a literal filesystem path and creates duplicate folder trees (`People &amp; Teams/` next to `People & Teams/`) — and pointing the agent at an already-affected note feeds the escaped path back in, so the damage self-reinforces (issue #1230).

## What

`normalizeLinkedContentPath` (`src/core/path/LinkedContentPath.ts`) now decodes exactly the entity set `escapePromptXmlAttribute` produces (`&amp;`, `&quot;`, `&lt;`, `&gt;`, `&#9;`, `&#10;`, `&#13;`) in a single pass, one level only:

- It is the shared canonicalization point — every entry into plugin state (session metadata decode, input-ledger normalization, `LinkedContentController`, `ConversationRepository`) flows through it, so structured/logged fields now always carry the raw vault path, and existing `path !== raw` handling flags escaped stored values with `needsMigration` so they self-heal on next save.
- One level only: `&amp;lt;` decodes to `&lt;`, not `<` — no double-decode loops.
- Escapes that decode to control characters (`&#9;`, `&#10;`, `&#13;`) are still rejected, so previously-invalid paths stay invalid.
- Raw paths containing a literal `&` are untouched.

## Why this approach

- Escaping stays at the render sites (correct for XML-attribute hygiene there); repair happens where the repo already canonicalizes and migrates paths, rather than adding a second decode at each storage site.
- The tradeoff is directionality: a vault entry genuinely named `People &amp; Teams` (which this very bug tends to create) normalizes to `People & Teams` after the fix. We chose unescape as the canonical direction because literal entity text in vault paths is almost exclusively bug fallout, mirroring how the module already canonicalizes separators and rejects escapes.
- Scope boundary, stated honestly: the rendered `<linked_content path="...">` attribute itself remains escaped (LLM consumers can still copy it); this PR guarantees the plugin never persists, re-serves, or re-renders a previously-ingested escaped value as structured data, and heals already-poisoned stored state. Prompt-format changes to make the attribute LLM-safe are a separate, larger discussion.

## Tests

`tests/unit/core/path/LinkedContentPath.test.ts` (colocated, mirroring `src/`):

- New `XML-escaped vault paths (issue #1230)` block: 8 cases — decode round trip (`&amp;`/`&lt;`/`&gt;`/`&quot;`), single-level decode guard, raw `&` paths unchanged, `needsMigration` flagging for canonical and legacy forms, control-character escapes still rejected.
- Red→green: all 8 new cases fail on unpatched `main` (`68b6266`) and pass after the patch; full file 30/30.

Validation on Windows (Node 24):

- `npx tsc --noEmit`: clean.
- `eslint` on both changed files: clean.
- Full unit+integration suite before/after, per-test-name comparison of the 29 suites that fail locally (all pre-existing Windows/environment failures, e.g. collab filesystem-boundary tests): 132 failed on `main` == 132 failed on this branch, name-by-name identical — 0 new failures, 0 resolved.
- `npm run build`: clean.

Fixes #1230


yetuge:
@randy-westman thanks — you're right, and I've narrowed the change (`76b3154f`).

**Your blast-radius analysis checks out.** The canonicalizer is fed ground-truth vault paths, not only persisted prompt text: `LinkedContentPickerSource.list()` (`TFile.path` / `TFolder.path`), `LinkedContentController.eligibleActiveFilePath()`, and `assertLinkedContentPath()` on rename/delete/create events. With the decode inside `normalizeLinkedContentPath`, a vault holding both `People & Teams` and `People &amp; Teams` collapses to one path, and `seenPaths` then drops the second real folder — exactly the picker affordance the recovery scenario needs. I reproduced it as a failing test before the change (`LinkedContentPickerSource` › "lists sibling real folders that differ only by entity-shaped names": two folders expected, one returned).

**What changed**
- `normalizeLinkedContentPath` is again a purely structural canonicalizer: entity-shaped substrings are ordinary name characters, so a real `People &amp; Teams` keeps its name everywhere ground truth is normalized (picker, active file, rename/delete events, settings pinning, conversation metadata writes). Five literal-name cases added as regressions.
- New `normalizeEscapedLinkedContentPath` (decode once, then canonicalize) is used only at the ingestion boundaries that carry provider/session text: `decodeLinkedContentPathFields` (canonical + legacy session metadata) and the input-ledger context normalization in `ConversationInputLedgerStorage`.
- The ledger read repairs, the ledger write keeps its previous strict contract. `save()` validates through the same normalizer, so decoding there would make a genuine entity-shaped vault path look like a migration candidate and throw; the read-only split avoids introducing that new failure mode. Test: "keeps the write contract for entity-shaped vault paths" asserts a literal `People &amp; Teams/note.md` still saves cleanly.

**The reported repro is still fixed.** Escaped canonical/legacy session metadata is still repaired with `needsMigration: true` (self-heal unchanged), and an escaped `context.linkedContent.path` in a legacy `.inputs.json` is still repaired on load with `needsMigration: true` (test: "repairs prompt-render escapes in the ledger Linked content path"). Escapes that decode to control characters are still rejected at the boundary.

**Residual, stated plainly.** Repair is a directionality decision wherever it applies: inside those two ingestion boundaries a *genuine* vault name that literally contains `&amp;` text is read as the decoded path. That is precisely why the decode now lives in the stores whose content is provider/session text instead of in the canonicalizer shared by every ground-truth caller. If you'd prefer an explicit trusted-vs-untrusted input distinction over a boundary split, that's a reasonable next step and I'm happy to go that way instead.

**Verification.** RED on the previous commit (`8cabbc1`): 5 literal-name cases plus the picker case failed. GREEN now: the three touched suites are 78/78; the full suite compared by test name against the previous commit shows no new failures — the only by-name differences are three git-plumbing integration cases under full-suite parallelism (`GitRepositoryService`, `NativeGitReviewRepository`) that pass 7/7 when the two suites are re-run in isolation, so they are environmental flakes rather than change-related. `tsc --noEmit` and eslint are clean, and the esbuild production build succeeds under the repo's expected Node 24.

randy-westman:
LGTM. Thank you!

YishenTu:
Thanks for narrowing the change in `76b3154f`. I read the explanation above and the subsequent LGTM. I understand that rewriting literal entity-shaped names on storage reads is an explicitly acknowledged tradeoff, and that changing the model-facing prompt format is outside this PR's stated scope. My objection is to that tradeoff at these particular boundaries, rather than to an overlooked picker regression.

The session metadata and input ledger are not exclusively provider-authored or XML-derived text: they also contain raw vault paths written directly by the plugin. For example, `InputController` builds `context.linkedContent.path` directly from `request.linkedContentPath`, and ledger writes preserve that literal value. A storage-read boundary therefore does not establish that the path was XML-escaped. This affects newly saved conversations too, not just historical records needing repair.

I ran a focused probe under Node 24.16.0 using this PR's ledger/path implementations through `ConversationInputLedgerStorage.save()` and `.load()`:

| Literal path saved | Path restored by this PR |
| --- | --- |
| `People &amp; Teams/note.md` | `People & Teams/note.md` |
| `Notes/&#10;x.md` | Linked content removed |
| `a &amp;lt; b.md` | `a &lt; b.md`, then `a < b.md` after another save/load cycle |

Current main preserves all of these. The first case demonstrates the acknowledged tradeoff; the latter cases show that it can also remove the reference or keep changing it across persistence cycles. Single-pass decoding within one call does not make the migration idempotent. The new write-contract test checks that saving succeeds, but not that loading restores the same path.

On the original issue: #1027 escapes paths at the XML prompt-rendering boundary. The fresh reproduction in https://github.com/YishenTu/claudian/issues/1230#issuecomment-5640585782 already shows the structured ledger path correctly stored as `Test & Notes.md`. This PR can repair a value that is genuinely escaped in storage, but that reproduction does not establish such a leak. For a correctly stored raw path, the model still receives the same escaped XML attribute, so the model-copying failure remains unchanged—as the PR's scope section acknowledges. I would not treat this as resolving #1230 without a reproduction establishing the storage-side failure.

I'd request changes: keep canonical stored paths literal, and only decode values whose XML-encoded provenance is established. If historical repair is needed, give it an explicit, one-time migration contract instead of applying it to every read. Addressing model interpretation at the prompt boundary can remain separate. The author already offered an explicit trusted-versus-untrusted distinction above; that is the direction I'd pursue, provided it tracks actual encoding provenance rather than merely which store a value came from.

Verification was limited to the focused storage probe and code/issue review; I did not run the full suite or a live-agent reproduction.

yetuge:
You're right on all three points, and I've replaced the approach rather than argue the tradeoff. `cdca8d04` drops the storage-side decoding entirely and moves the fix to the prompt boundary you pointed at. The branch is now four files: the two prompts that document these tags, and their tests.

**I reproduced your probe before changing anything.** All three cases behave exactly as you described on the previous head:

| Literal path saved | Restored after the decode |
| --- | --- |
| `People &amp; Teams/note.md` | `People & Teams/note.md` |
| `Notes/&#10;x.md` | reference removed |
| `a &amp;lt; b.md` | `a &lt; b.md`, then `a < b.md` after another cycle |

The third one is the one that settled it for me: single-pass decoding within a call does not make the migration idempotent, so "repair on read" was a transformation that could keep changing a value that was never wrong.

**Your point about provenance is the part I had wrong.** I reasoned from the store rather than from the writer. `ConversationInputLedgerStorage` writes from `request.linkedContentPath`, which comes from the picker or an `@` mention — a real Vault path that never passed through `escapePromptXmlAttribute`. So the ledger does not establish XML-escaped provenance, which means it is the wrong boundary for a heuristic the function cannot actually verify.

**The actual gap is in the prompts.** `escapePromptXmlAttribute` is applied at every render site, and both system prompts document the tag shapes and `<![CDATA[...]]>` — but neither ever says the `path=` attribute is escaped. The agent receives `path="People &amp; Teams"` with nothing telling it that `&amp;` is an encoded `&`, so treating it as literal path text is the reasonable reading of what it was given. Both prompts now name the escape set, say to unescape before using a value as a path, and say never to write an escaped form to disk.

The tests pin the documented set to the renderer's own output — they call `escapePromptXmlAttribute` and require every entity it emits to be named in `## User Message Context` and `## Input Context`. If an escape is ever added to the renderer, the suite fails until the prompts document it, so this cannot drift apart again silently.

**Verification**
- `tests/unit/core/prompt/` — 47 passed. Both new cases fail on the pre-fix prompts (confirmed by stashing only the prompt changes) and pass after.
- `tests/unit/core` + `tests/unit/utils` — 1674 passed / 4 failed against a 1671 passed / 4 failed baseline on `main`; the 4 are Windows `EPERM ... symlink` failures in `VaultFileAdapter`, identical both sides.
- `tsc --noEmit` — 501 errors on `main`, 501 with the change, all in the collab test trees and none in the touched files.

Not covered, so you know the limit of what I checked: I did not run a live agent against a real folder containing `&`. This is a prompt-level instruction, so it guides the model rather than forcing the path, and a model that ignores it can still reproduce the escaped form. If you would rather have a guarantee than a strong instruction, the durable version is to throw on assignment into the structured field so it cannot hold an escaped value at all — I left that out because it is a storage-contract change and you asked for the prompt boundary, but I'm happy to do it as a follow-up if you want the failure mode closed by construction.

PR #1341, opened 2026-09-15
feat(mentions): allow @ to reference files outside the vault

rawlencecn:
## Why

Today `@` mentions can only reference files inside the vault, because every
suggestion comes from Obsidian's Vault API (`app.vault.getFiles()` /
`getAllLoadedFiles()`), which is scoped to the vault root. Anyone whose working
material lives outside the vault — a code repo, notes in `~/Documents`, a
sibling folder — cannot reference it with `@`, even though the agent's working
directory is the vault and it can read external paths when granted access. This
closes that gap with an explicit, opt-in capability.

## What

- **New setting "Allow @ mentions outside the vault"** (off by default),
  localized across all bundled locales. When off, nothing is registered and
  behavior is unchanged.
- **`FilesystemMentionSource`** (new shared composer-dropdown source) browses
  the filesystem via `readdir`:
  - Engages only for unambiguous path-shaped queries: absolute (`/…`), home
    (`~/…`), Windows drive/UNC (`C:\…`, `\\…`), and vault-relative (`./…`,
    `../…`, resolved against the vault root like a terminal resolves against the
    cwd).
  - Lists directories first, hides dotfiles unless the query starts with `.`.
  - Files and directories are both directly selectable and insert an absolute
    path; directories keep a trailing slash. This matches the in-vault folder
    mention UX — one Enter references a folder, deeper browsing is done by
    continuing to type the path.
- **Source hand-off**: the vault `MentionSource` now declines path-shaped
  queries so the two sources never contend for the same `@`, and any space
  after a path closes the mention (no re-triggering after a completed path).
- **Agent access**:
  - Claude — the mentioned file's directory is added to
    `options.additionalDirectories`.
  - Codex — the directory is added to the workspace-write sandbox
    `writableRoots`.
  - Both resolve relative mentions against the vault working directory, and the
    main-agent system prompt documents the new path shapes.

## Why this approach

- **Reuses the existing dropdown source abstraction** rather than bolting file
  browsing onto the vault source, keeping vault and filesystem concerns
  separate and each source's `match()` authoritative for its own query shapes.
- **Path-shape gating + explicit setting** means zero behavior change and zero
  cost when disabled, and no risk of shadowing ordinary `@name` vault lookups
  when enabled.
- **Directory-as-selectable (not drill-down)** was chosen after iterating on a
  drill-down design; matching the vault folder UX proved less surprising.
- **Pure path helpers are isolated from Node** (`src/utils/externalMention.ts`,
  the source core) so they are unit-testable with in-memory fakes; only a thin
  adapter touches `fs`. Provider wiring stays within each provider's own
  execution code — no shared feature code reaches into provider internals.
- Considered alternatives: a full ripgrep/git-backed index (as in Claude Code)
  was rejected as too heavy for the Obsidian bundle-size constraint and
  unnecessary for interactive per-directory browsing.

## Validation

- `npm run typecheck` — clean.
- `npm run lint` — clean.
- Unit tests added/updated and passing:
  - `tests/unit/utils/externalMention.test.ts` — path detection, deepest-dir
    resolution, `~`/relative expansion, directory extraction for authorization.
  - `tests/unit/shared/composer-dropdown/FilesystemMentionSource.test.ts` —
    matching, listing/sorting/dotfiles, relative browsing, directory/file
    selection, space-closes-mention.
  - `tests/unit/shared/composer-dropdown/MentionSource.test.ts` — vault source
    declines external path queries.
  - Existing dropdown/FileContextManager/system-prompt/settings/locale suites
    remain green.
- Manually verified in Obsidian (Claude + Codex): `@/abs`, `@~/…`, `@./`,
  `@../` browse and reference correctly; folder mention is a single Enter;
  dropdown closes after a completed path; disabled setting leaves `@` unchanged.

## Impact

- Off by default; no change for users who don't enable it.
- **Limitation**: agent-side access is wired for Claude and Codex. Other
  providers (Grok, Opencode, Pi) will accept the inserted absolute path but do
  not yet register it for read access — a natural follow-up.
- Codex grants read+write to the directory (workspace-write sandbox has no
  read-only-append granularity); Claude's `additionalDirectories` is read scope.
- No new production dependencies.


YishenTu:
i think we should restrict the @mention to a certain scope, a simple toggle is not enough.

